PR #80, opened 2020-06-27
Fix typo in error message

halgari:
Missing the `$` that will let me see why my modlist won't load


Noggog:
Gotta figure out my notifications, apparently.  Didn't even notice this until just now.  Still isn't showing up as an event of interest in my bell area, weird.

PR #126, opened 2020-10-16
Added IsDeleted

ghost:
Added IsDeleted to check if a record is flagged as deleted or not. Works for all games.


Noggog:
Looking good!  While we're here, can we make one tangential improvement?  

Right now everyone is listing the raw 0x0 value everywhere, which just happens to line up with each other.  Would be cool to have `OblivionMajorRecordFlags` and `SkyrimMajorRecordFlags` point to and use directly the Constants values you just added straight up, so there's no chance for typo "desync".   Both for Deleted and Compressed

The namespaces might get a bit weird, since most of the libraries have a `Constants` file.. so you may need to be explicit and do `Mutagen.Bethesda.Internals.Constants.CompressedFlag`

Issue #140, opened 2020-11-04
MajorRecord.FormKeyComparer


Noggog:
Similar to FormKey.LoadOrderComparer or FormKey.AlphabeticalComparer. MajorRecord should have a MajorRecord.FormKeyComparer that returns `IEqualityComparer<IMajorRecordCommonGetter>` that does equality/hash based off FormKeys

Could also add a generic <T> variant (scoped to IMajorRecordCommonGetter) that also ensures the types are desirable

PR #141, opened 2020-11-04
MajorRecord.FormKeyComparer for key-based Equality comparison.

Deigue:
For addressing Issue #140


Mutagen.Bethesda.Core/Records/MajorRecord.cs
```diff
@@ -65,6 +65,28 @@ object IDuplicatable.Duplicate(Func<FormKey> getNextFormKey, IList<(IMajorRecord
         {
             return this.Duplicate(getNextFormKey, duplicatedRecordTracker);
         }
+
+        #region Comparers
+        public IEqualityComparer<IMajorRecordCommonGetter> FormKeyComparer()
```
Noggog: Let's make this static, too.  Right now we'd need a major record instance to get a comparer

Deigue:
Ah good catch, updated it ... not sure how I missed the main method that I will be using to be a singleton lol.

PR #155, opened 2020-11-18
API Enhancements for Disabling Records.

Deigue:
Made a bunch of API enhancements for ease of disabling records, which can be a quite common use-case from perspective of patcher devs.

- Created IPlaced.cs for partial user interface implementation, custom override of the inherited Disable() method
- Moved over Disable() functionality to IMajorCommonRecord, as it is more commonly exposed as a TSetter generic type in other areas such as OverrideMixIns.
- Implemented helper/convenience methods to quickly override and disable things in appropriate places

Targeted it towards the `dev` branch to get a cleaner/strict compare, since it was built on top of that rather than the current `release` branch.


Noggog:
Looking good overall!

Few tweaks I'd suggest
### This is one area where I think the convenience methods actually harm the API

For example, typical usage will inevitably be:
```cs
state.PatchMod.Npcs.OverrideAndDisable(npcGetter);
// or
placedObjectContext.OverrideAndDisable(state.PatchMod);
```

This doesn't lead cleanly into the IPlaced customized override, though.  Users will see `placedObjectContext.OverrideAndDisable` and see zero alternatives in the intellisense.   It would be ideal to set it up so that the customized calls for IPlaced slip in naturally as an alternative.

If we nuke the convenience methods, we instead get this as the typical API, which is still decently clean:
```cs
state.PatchMod.Npcs.GetOrAddAsOverride(npcGetter).Disable();

placedObjectContext.GetOrAddAsOverride(state.PatchMod).Disable();
```
From this setup, the customize setup for IPlaced shows up naturally as an alternative:
```cs
placedObjectContext.GetOrAddAsOverride(state.PatchMod).Disable(DisableType.SafeDisable);
```
This call would show as the 2/2 alternative in intellisense.

**TLDR:**  I think we should actually delete the convenience extension methods

### Interface default methods don't actually hook too cleanly
What I mean is that a standalone default implementation in an interface can be called only if the variable is an interface:
```cs
PlacedObject straightObj = ...;
IPlaced interf = straightObj;
// works
interf.Disable();
// breaks
straightObj.Disable();
```
Weird...

Additionally:
```cs
PlacedObject obj = ...;
IPlaced placedInterf = obj;
IMajorRecord majorInterf = obj;
placedInterf.Disable(); // Calls IPlaced's disable, with SafeDisable logic
majorInterf.Disable() // Calls IMajorRecord's disable, with JustInitiallyDisabled logic
```
This.. is dangerous.  I think this reason alone I don't like default interfaces much.

TLDR:  I wouldn't use default interfaces.  Just declare the method without a body normally, and implement in MajorRecord.cs, I'd say.

### The current setup is an API clash
This will cause the compiler to freak and throw an "ambiguous" error:
```cs
IPlaced placed = ...;
placed.Disable();
```
It won't know whether to call the MajorRecord.Disable(), or the IPlaced.Disable(), since the IPlacedDisabled doesn't require parameters but is a completely separate call, it's ambiguous which is the intended target

To solve this, I would:
- Make IPlaced.Disable()'s parameter non-optional, so that it is distinct and removed the ambiguity
- Make the method we added to MajorRecord.cs virtual
- Override in PlacedObject/Npc/Trap and call `IPlaced.Dispose(SafeDisable)`, which is the default we want.

This way, a call to Disable() is not ambiguous, and also hooks in and calls the SafeDisable logic if it's IPlaced.

Deigue:
- I am fine with that, Its essentially a wrapper for two calls, which can easily made patcher-wise client side if need be.
Yes, it has potential to hide API .. rather if you typed "Override" I would assume it includes "GetOrAddAsOverride" in the listbox, but in any case it can be tempting and confusing. Will nuke these 
- Oh yea, you made a comment amount this oddity earlier but it slipped my mind. For MajorRecord this is trivial, ill fix that for sure. For IPlaced it does mean duplicated code across APlacedTrap/PlacedObject/PlacedNpc (This was the core reason why I did it to be honest, I wanted to avoid unnecessary duplication of logic if possible, but if it causes a drastic effect like it not being available in the wrong context, it defeats the whole purpose. So is the duplication a necessary evil here?
The other option is a abstract class covering all 3 ... but definitely super overkill for just one method right?
- Nice, yes ill go for it that approach.

Deigue:
Should cover/address all the above points ... let me know if anything is missed.
Might need to look on per commit basis to filter out Generated changes if needed.

PR #178, opened 2021-02-28
Sqlite formkey allocator

mrudat:


Mutagen.Bethesda.Core/Persistance/SQLiteFormKeyAllocator.cs
```diff
@@ -0,0 +1,166 @@
...
+
+        private readonly DbSet<SQLiteFormKeyAllocatorFormIDRecord> FormIDs;
+
+        private readonly DbSet<SQLiteFormKeyAllocatorPatcherRecord> Patchers;
+
+        public SQLiteFormKeyAllocator(IMod mod, string patcherName, string dbPath)
+        {
+            Mod = mod;
+
+            Console.WriteLine($"Maintaining FormID allocation database in {dbPath}");
```
Noggog: Don't want to do this in a class that's in a common lib.  Not sure how people will be using this class or if they'll want it to print

The user of this class should be the one to print if they want

Mutagen.Bethesda.Core/Persistance/SQLiteFormKeyAllocator.cs
```diff
@@ -0,0 +1,166 @@
...
+        {
+            if (editorID == null) return GetNextFormKey();
+
+            lock (AllocatedEditorIDs)
+                if(!AllocatedEditorIDs.Add(editorID))
+                    throw new ConstraintException($"Attempted to allocate a duiplicate unique FormKey for {editorID}");
+
+            SQLiteFormKeyAllocatorFormIDRecord? rec;
+
+            lock (this)
```
Noggog: I like the locks for the multithread safety, but I think it's subtlety unsafe to `lock (this)`, though.   It'll work just fine until a user of this class does their own lock on the object, at which point it'll fight with these locks internally in a potentially undesirable way.

If I want to `lock (this)`, instead I'll just make a `private readonly object _lockObj = new object();` and lock on that, just so that it's technically a different object than the SQLiteFormKeyAllocator i'm giving the end users.  It lets them write locks on the SQLiteFormKeyAllocator without any potential for odd interactions

Mutagen.Bethesda.Core/Persistance/SQLiteFormKeyAllocator.cs
```diff
@@ -0,0 +1,166 @@
...
+
+                return new FormKey(Mod.ModKey, candidateFormID);
+            }
+        }
+
+        public FormKey GetNextFormKey(string? editorID)
+        {
+            if (editorID == null) return GetNextFormKey();
+
+            lock (AllocatedEditorIDs)
```
Noggog: II think we can merge all these locks in this statement to just one overarching one on `_lockObj`.  Don't think the cost of reacquire the lock 3 times is worth the ability for other queries to do work between the locked statements.  Most of the work in between locking is short enough to just do it within the one lock, I think?

Mutagen.Bethesda.Core/Persistance/SQLiteFormKeyAllocator.cs
```diff
@@ -0,0 +1,166 @@
...
+                if(!AllocatedEditorIDs.Add(editorID))
+                    throw new ConstraintException($"Attempted to allocate a duiplicate unique FormKey for {editorID}");
```
Noggog: "duplicate" typo

Mutagen.Bethesda.Core/Persistance/SQLiteFormKeyAllocator.cs
```diff
@@ -0,0 +1,166 @@
...
+        }
+
+        private object GetPatcherName(uint patcherID)
+        {
+            lock (this)
+            {
+                // should be $"select PatcherName from Patchers where PatcherID = {patcherID}"
+                return Patchers.Where(r => r.PatcherID == patcherID).Select(r => r.PatcherName);
+            }
+        }
```
Noggog: Don't like the `object` return here, which actually seems to be an `IQueryable<string>` that we're returning.  Not sure how that'll print in the exception message this function is used for?  I think something like this will be better:
```
private string GetPatcherName(uint patcherID)
{
    lock (this)
    {
        // should be $"select PatcherName from Patchers where PatcherID = {patcherID}"
        return Patchers.Where(r => r.PatcherID == patcherID).FirstOrDefault()?.PatcherName ?? "Unknown Patcher";
    }
}
```

Noggog:
Thanks for the PR!  Looking super cool
Added some comments with a few small tweaks.

Aside from those, I have one other thought:
I was stewing on the idea/inclusion of `string patcherName` as an input to make a `SQLiteFormKeyAllocator`.  It's maybe something we should consider offering an alternative that doesn't take it.  For example, it makes perfect sense in the context of Synthesis, which is constructing a single mod (`Synthesis.esp`) from several separate runs (different patchers).  As such, the extra inclusion of `string patcherName` is extra functionality helps avoid collisions cross-patcher, which is a super nice feature to have.

But I think for more simplistic non-synthesis users, though, this parameter is going to be confusing and seen as unnecessary.  The user already has created a mod object, which has its own ModKey (MyCoolMod.esp) which they're giving the allocator.  Now they have to give it another string in addition?  For them, the concept of `patcherName` isn't a concept, so do they just give any string at all for this parameter?   Would be better to just not require them to have to give one.  They just want an allocator for their mod, and for it to give FormKeys based on saved EDIDs -> no patcherName concepts at all.

Perhaps we could offer this up by allowing patcherName to be null?  `string?`
This way we can offer a ctor that doesn't take patcherName, for the more simplistic non-Synthesis users
I think it could do similar queries, it just wouldn't do that extra patcherName collision check

Either this can just be made an optional parameter, or maybe eventually we just offer two allocator classes /w different flavors?  Either way would probably be fine

Mutagen.Bethesda.Core/Persistance/SQLiteFormKeyAllocator.cs
mrudat: tyop fixed.

Mutagen.Bethesda.Core/Persistance/SQLiteFormKeyAllocator.cs
mrudat: Oops. Fixed.

mrudat:
> Either this can just be made an optional parameter, or maybe eventually we just offer two allocator classes /w different flavors? Either way would probably be fine

Hmm. I added an optional parameter (moved the optional parameter to the last position).

I'll see about a distinct class, though, I think. The code for not caring about multiple distinct writers sharing a single output file seems like it would be significantly simpler than having two or more writers issuing FormIDs off the top of my head.

Mutagen.Bethesda.Core/Persistance/SQLiteFormKeyAllocator.cs
mrudat: Hmm. the main reason for `lock (this)` is that the recommendation is to lock the `DbContext` object, which represents the open handle to the database; it just happens that we're holding open a database connection for the entire time that `SQLiteFormKeyAllocator` exists, so I thought that I'd be lazy and make SQLiteFormKeyAllocator also the database connection object, so I don't need to (for example) forward `Dispose`.

I'm not sure if it would be better to just lock a new object, or to go back to a separate DbContext object, and lock that instead.

Noggog: If we're planning on exposing `SQLiteFormKeyAllocator`, I think it'd be best to make the DbContext another internal non-public class, that we use inside `SQLiteFormKeyAllocator`.
One reason would be that if `SQLiteFormKeyAllocator` IS the DbContext itself, then people using `SQLiteFormKeyAllocator` also have access to all the public things DbContext has -> A whole lot of functions we don't want the outside user mucking with

If the DbContext is an internal class we're wrapping, we can also then lock it pretty easily like you are mentioning, too.

on that note, I think `class SQLiteFormKeyAllocator` itself needs to be public, so people can use it?

Noggog:
> I added an optional parameter (moved the optional parameter to the last position).

Yeah, optional parameter is fine, I think.  Only make a 2nd class if the differences are so large it'd be easier to code that way.  Probably not

Mutagen.Bethesda.Core/Persistance/SQLiteFormKeyAllocator.cs
mrudat: I've pulled out the DbContext to its own object.

Mutagen.Bethesda.Core/Persistance/SQLiteFormKeyAllocator.cs
mrudat: I had to lock the context in `GetNextFormKey()` in order to get it to work, but that should hopefully go faster.

PR #179, opened 2021-03-02
updated TextFileFormKeyAllocator

mrudat:
a new version of TextFileFormKeyAllocator that should have the same behaviour as SQLiteFormKeyAllocator


Noggog:
I think for this one, I'd like to actually keep TextFileFormKeyAllocatorV2 its own separate additional entity (different name, ofc).  As I stew on things, I'm starting to see the added feature of `string patcherName` as not really core to the `IFormKeyAllocator` concepts.  It's more of an additional extra ask/feature for special snowflakes like Synthesis specifically.  Almost so much so that it wouldn't be crazy to see it as a specialized allocator for Synthesis IN Synthesis' repo?  But I can also see rare other users might find it useful so we could keep in Mutagen, but it's definitely not the typical use case 90%+ of people will use.

I think it's good to keep what is usually the more typical use case straight up, as TextFileFormKeyAllocator currently does.
I'm making a mod -> and id like to track the formkeys i made last time.  No concept of other mods or past partial jobs making things complex, etc.

For example, with TextFileFormKeyAllocator, we can track a file like this:
```
ModA.esp (file)
   EDID1 -> 800
   EDID2 -> 801
   EDID3 -> 802
```
   
And the system can tell us that EDID2 was associated with 802 (and presumedly will result in 000802:ModA.esp)
And that the next available FormID would be 803

Simple setup.  TextFileFormKeyAllocator just uses a file to track used EDIDs associated with a mod.  No fancy extras


--------------------

I think we can offer other advanced use case functionalities like an additional `string patcherName` key, but they should be offered on the side in addition to.  They can either be their own implementations, or utilize composition to offer the functionality (eg, contain TextFileFormKeyAllocators internally).

For example, considering we have:
```
Patcher1.esp (file)
   EDID1 -> 800
   EDID2 -> 801
   
Patcher2.esp (file)
   EDID3 -> 802
   EDID4 -> 803
```
   
Let's say synthesis is currently patching Patcher2, it would give a SynthesisTextFileFormKeyAllocator a folder containing all these files ^ to look at, while passing Patcher2.esp as the "active" file.
The fancy class would load everything up into several TextFileFormKeyAllocators internally.

Requests to the SynthesisTextFileFormKeyAllocator would just make the requests to the appropriate Patcher2 sub-allocator, while first checking with the others for collisions from past patchers we want to block.  Collision happens on another patcher's file?  Throw an exception with the patcher name.  (Maybe we add `HasKeyFor(string edid)` to `IFormKeyAllocator` to power this)

this keeps most of the logic within TextFileFormKeyAllocator, and the SynthesisTextFileFormKeyAllocator class would just be the coordination of how to utilize multiple TextFileFormKeyAllocator to cross reference for extra EDID query rejections for the patcher-over-multiple-runs concepts.

If we wanted to, or if we found that actually using several TextFileFormKeyAllocator internally was slow, we could do a custom implementation under the hood for optimization.

But I think these are extra advanced requirements we're adding here, and I think trying to artificially meld it with what could be a simplistic TextFileFormKeyAllocator might be a mistake.   For example, now it has two write methods, each of which can throw if called for the wrong reason.   I think instead we can just rename the class you made to something else, and offer it as a separate feature/choice?


--------------------
The reason it makes a bit more sense for the Sqlite side to be one big thing is that that way we can have one single database everyone can bounce off of with some fancy indexing.  So in Sqlite's situation is sort of more worth it to have one superclass? as we can make a more optimized database internally?  Still not 100% sure.

mrudat:
A ISynthesisFormKeyAllocator for the extra functionality?

SQLiteFormKeyAllocator just happens to support both?

Noggog:
Yeah, something like that.   Not sure on the name.  Maybe IMultiphaseFormKeyAllocator?  With maybe "phaseName" as the parameter wording?

Essentially it's a normal allocator, but with the extra concept that it helps construct a single mod over the course of multiple phases/proccesses.  The phases need tracking on-disk, so that the 2nd phase doesn't request the same EDID as the 1st phase from earlier.   If it was done all at once in-memory and in one process, this wouldn't be needed.  It's only for when a process ends, and we're starting another to keep working on the same mod -> each process is a new phase that needs a nickname

Noggog:
Also, it could come as another pattern
Instead of giving each phase a nickname, we could go the route of giving it two paths:
- One to the file/db of all the mappings, as per usual
- One file/db of the edids requested so far.  Doesn't matter how many phases contributed to this file, each one would just add to it over the course of the work.   After the whole thing is done, this file can be tossed.  The first file is the complete mapping file of all EDIDS ever requested.  This one was just to track the current "run"

I guess the downside of this would be any collisions wouldn't be able to inform the user which phase was the one that previously requested the EDID causing the collision.  So maybe that's an important enough feature to warrant the phase name tracking systems we have now.  Figured id bring it up as musing food for thought, though.

mrudat:
If you don't want to name the phases, you could keep track of the number of phases so far, and offer `$"phase{n+1}"` as the name of the next phase? Perhaps by calling the allocator creation in the right way?

I think that the better error messages are probably worth the extra memory/time spent tracking which phase first allocated which FormKey.

Is it useful to have a distinct IPersistentFormKeyAllocator that presents a Save/Load interface?

Then again, I'm not sure how useful it is to have 4 classes and 3 interfaces...

I'm wondering if it would be the Right Thing to have:
* StateDirectory\ATarget.esp.sqlite <-- SQLiteFormKeyAllocator
* StateDirectory\TheTarget.esp.txt <-- TextFileFormKeyAllocator
* StateDirectory\Shared.esp\phase1.txt <-- TextFileSharedFormKeyAllocator
* StateDirectory\Shared.esp\phase2.txt

The interface for all persistent FormKeyAllocators would be something like:
```C#
formKeyAllocator.Read("...\StateDirectory");

formKeyAllocator.Write("...\StateDirectory");
```

Noggog:
I think the allocators should load on creation, and save either on request, or on disposal (can maybe have them implement IDisposable for this).  Makes the api easy:
```
using var alloc = new TextFileFormKeyAllocator(path, ...);
var formKey = alloc.GetNextFormKey("someEdid");
// Request automatically saved as alloc is cleaned up
```

I think the supersystem that's combining all the alternative formkey allocator types is probably forecasting what the user will want a bit too much.   I think typical users will want to just use a single allocator of their preferred type syncing to a single file, and that's it.  Anyone that wants a super custom setup (especially /w mixed allocator types) probably has a lot of other requirements too.  Might as well just let them combine them how they want themselves since they know their requirements.

I think things are looking pretty good as-is.  Anything else you think we need?  We can obviously always tweak/improve a bit more as we see things later

mrudat:
Would it be the Right Thing to accept an optional CancellationToken (given Synthesis uses them already), and have Dispose save if all goes well, and the save abort if cancelled?

Not sure if that's something that's reasonable to do for something that's supposed to run in-process, but it might work?

It should make the simple case easy, and the complicated case possible.

Noggog:
Nah, I think the cancellation token isn't really needed.  You can add them to literally EVERY function, if you wanted to, as literally any job/function has the potential to be cancelled early.  But I think it's only appropriate to put it on places you have a specific use case in mind, otherwise it just gets to be out of control.  I don't think saving out persisted FormKey mappings is a scenario where someone would want to cancel their save as it's occuring.  If someone requests the ability to for a weird obscure reason, we could just add it then without breaking the API at all

On the topic of saving, though, are you totally sold on having the dedicated Save() call?  I think it's maybe fine to offer as a voluntary "early" save (eg, can call it every 5min, to just save the current state out)

But I think Save should also automatically happen in Dispose, as well.  Again, so this API is viable (which i find to be the cleanest):
`using var allocator = new SomeAllocator()`
no one needs to remember to call Save, as it's automatically remembered/handled on disposal, which is automatically done when allocator goes out of scope.

mrudat:
After having [attempted to wire it into Synthesis](https://github.com/Mutagen-Modding/Synthesis/pull/172), I can see the appeal of saving on Dispose. It's mostly that the documentation for Disposing of database-related things suggests that you should throw the changes away if anything goes wrong.

ie.
```C#
var allocator = new Thing();
try {
    // do something tricky here.
    allocator.Commit();
}
catch (Exception e) {
    allocator.Rollback();
}
finally {
    allocator.Dispose(); // Could Rollback, or could Commit, in which case you can omit one of the previous calls.
}
```

Noggog:
Ah yeah.  Hmm.

I think we can offer `Save`/`Commit` for the Sqlite one, in which case users that want a specific behavior can do code like you posted above.  What are they gonna do, try to commit again, or something?  I guess they can try.

Most people (and all the other allocators) would just utilize the Dispose setup?   And for the sqlite one it can be equivalent to the above code, where it rolls back/disposes/etc.  We could then also just rethrow the exception, so they're aware something went wrong.  They just wouldn';t have the choice NOT to roll back, it would roll back and then rethrow.  But that's probably the typical use case anyway

Mutagen.Bethesda.Core/Persistance/ISharedFormKeyAllocator.cs
```diff
@@ -0,0 +1,26 @@
...
+        {
+            PatcherName = patcherName;
+        }
+
+    }
+
+    public interface ISharedFormKeyAllocator : IPersistentFormKeyAllocator
+    {
+
+    }
```
Noggog, line 25: Don't know that we need a new interface defined here, if it's not exposing anything new.

In fact, I don't know that we need the abstract class either, as it's not really coalescing any common logic that would be reused by a 2nd shared allocator.  Not sure that just storing PatcherName is enough to warrant the definition.  But if we change our minds and want to coalesce some common logic in the future, we can always define this abstract class then, without breaking any backwards compatibilities.  Since we can do it later when it makes sense, we might as well keep it simple for now until we need the complexity.

Mutagen.Bethesda.Core/Persistance/TextFileSharedFormKeyAllocator.cs
```diff
@@ -0,0 +1,138 @@
...
+    public class TextFileSharedFormKeyAllocator : BaseSharedFormKeyAllocator
+    {
+        private readonly Dictionary<string, (string patcherName, FormKey formKey)> _cache = new();
+
+        private readonly HashSet<uint> FormIDSet = new();
+
+        public TextFileSharedFormKeyAllocator(IMod mod, DirectoryPath saveFolder) : base(mod, saveFolder)
+        {
+            Load();
+        }
```
Noggog, line 25: I think this ctor /w no patcher name only made sense when TextFileFormKeyAllocator was trying to do everything.  TextFileSharedFormKeyAllocator should always require an "active" patcher name as input, I think.  It doesn't make sense to use this system if you don't have an active patcher in mind

Mutagen.Bethesda.Core/Persistance/TextFileSharedFormKeyAllocator.cs
```diff
@@ -0,0 +1,138 @@
...
+
+        public override void Rollback()
+        {
+            lock (_cache)
+            {
+                _cache.Clear();
+                FormIDSet.Clear();
+                Load();
+            }
+        }
```
Noggog, line 118: I think we'd want to roll back the Mod.NextFormID to its original value, too?

Mutagen.Bethesda.Core/Persistance/TextFileFormKeyAllocator.cs
```diff
@@ -72,39 +36,75 @@ public FormKey GetNextFormKey()
...
-                        var edid = m.EditorID;
-                        if (edid == null) return TryGet<KeyValuePair<string, FormKey>>.Failure;
-                        return TryGet<KeyValuePair<string, FormKey>>.Succeed(
-                            new KeyValuePair<string, FormKey>(edid, m.FormKey));
-                    }));
+            lock (Mod)
+            {
+                _cache.Clear();
+                Load();
+            }
```
Noggog, line 107: I think we'd want to roll back the Mod.NextFormID to its original value, too?

Mutagen.Bethesda.Core/Persistance/IPersistentFormKeyAllocator.cs
```diff
@@ -0,0 +1,60 @@
...
+        protected IPath SaveLocation;
+
+        public bool CommitOnDispose = true;
+
+        private bool _disposed = false;
+
+        protected BasePersistentFormKeyAllocator(IMod mod, IPath saveLocation) : base(mod)
+        {
+            this.SaveLocation = saveLocation;
+        }
```
Noggog, line 17: I think `string` is more appropriate here for `saveLocation`.   It's a weird subtle topic, that i'm still simmering on myself... but currently I think `string` is currently most appropriate when we're just passing pathing information.  `IPath` has too many other concepts attached to it -> `bool Exists` for example.

I think IPath/FileInfo/DirectoryInfo/FilePath/DirectoryPath are more heavyweight objects that have a lot more state to them than needed, typically.  (FilePath/DirectoryPath are my personal wraps around FileInfo/DirectoryInfo /w preferred API, but im still unsure about them).  For just passing/manipulating paths, it's arguable that string itself is good enough.   And/or a special struct could be made that doesn't store any state.  WJ has these concepts of stateless path structs
https://github.com/wabbajack-tools/wabbajack/blob/master/Wabbajack.Common/Paths/AbsolutePath.cs
Note that exists doesn't store a state, it actively rechecks every call, which is different than FileInfo/FilePath

Here's a stack overflow about how FileInfo/DirectoryInfo can be deceiving as they store state in addition to pathing info:
https://stackoverflow.com/a/51573245/7971470

TLDR:  Probably just pass paths around as `strings`, for now.  Seems to be the C# standard, and I haven't figured out my truly preferred alternative yet to commit to anything else

Mutagen.Bethesda.Core/Persistance/IPersistentFormKeyAllocator.cs
```diff
@@ -0,0 +1,60 @@
...
+        public void Dispose() => Dispose(true);
+
+        /*
+        public void Dispose()
+        {
+            // Do not change this code. Put cleanup code in 'Dispose(bool disposing)' method
+            Dispose(disposing: true);
+            GC.SuppressFinalize(this);
+        }
+        */
```
Noggog, line 48: I usually like just nuking commented out code.  Git's always there if we change our mind want to retrieve them from the past

Mutagen.Bethesda.Core/Persistance/TextFileSharedFormKeyAllocator.cs
```diff
@@ -0,0 +1,138 @@
...
+
+        public TextFileSharedFormKeyAllocator(IMod mod, DirectoryPath saveFolder, string patcherName) : base(mod, saveFolder, patcherName)
+        {
+            Load();
+        }
+
+        private void Load()
+        {
+            foreach (var file in Directory.GetFiles(SaveLocation.Path, "*.txt"))
+                ReadFile(file, Path.GetFileName(file)[0..^4]);
```
Noggog, line 35: Swap this to Path.GetFileNameWithoutExtension

PR #189, opened 2021-03-23
Add MutagenMerger.Pex

erri120:


erri120:
Will re-open after fixing an issue

Noggog:
Awesome, very clear polished stuff!  

I'd like adjust a few things to the existing patterns a bit:

## One Point of Access
SkyrimMod, for example, which has no (public) helper classes:
	- Reading a skyrim mod?  `SkyrimMod.CreateFromBinary(path)`
	- Reading a skyrim mod for overlay?  `SkyrimMod.CreateFromBinaryOverlay(path)`
	- New skyrim mod? `new SkyrimMod(..)`
	- Writing a skyrim mod? `skyrimMod.WriteToBinary(path)`
	
For Pex, right now there's two namespaces and two different classes to go to, depending on if the user wants to import or create new.  
    - `Mutagen.Bethesda.Pex.PexParser` if they want to import. 
    - `Mutagen.Bethesda.Pex.DataTypes.PexFile` if they want to make new.  

Doesn't really add much power to make people go two places for very similar goals.  I think PexParser's factory can be absorbed into PexFile, which would mimic SkyrimMod's construction patterns and be a nice one stop shop.  The write function can exist in a static extension class, that's fine, since that separation is invisible to the user
	
## Derivative Fields Exposed in Interface
The current API is very true to the disk structure.  Not necessarily a bad thing if you want extreme control, but Mutagen tends to lean towards abstraction for the benefit of the user.  I think there's a few concepts we could abstract away to improve the public usability people see in Mutagen.  

The simple example is the exposure of `InstructionCount`.   `InstructionCount` only exists as a field someone can forget to set properly when modifying `LineNumbers`.  Storing it as an exposed variable just opens up the potential for desync between it and the true value `LineNumbers.Count`.   Also, it exposes some potential confusion for the users.  Do they check `InstructionCount` or `LineNumbers.Count`?  Do they always agree?  Can they differ?  Which is the right one to use?  

There's a few `Num` fields that are similar.  They're derivative fields that shouldn't be exposed, as they're implicit in the paired collection's count.

----------

Additionally, I think the concept of IStringTable and public index members can be hidden and absorbed to being an internal concept.  There's lots of places Mutagen hides away busywork details that don't add much.   A decent equivalent is Mutagen's current encapsulation of the Strings file concepts.  People don't care that an Npc's name is fulfilled by a strings file for Skyrim vs Oblivion.  They just want to see Npc's `string Name;` field and use it.  The fact that technically `FULL` has an index to be looked up in the strings folder is an implementation detail that's not too interesting to the user.

Similarly, I don't think anyone cares about having exposure to the Pex indices.  That's just a storage detail of the pex format on-disk.  `DebugFunction` HAS a FunctionName, which is what they're after, it's just achieved via combining index + the strings file.   We could just offer `string FunctionName { get; set; }` and people would be much happier.

Additionally, and more importantly, I think it's actually safer and more robust of an exposed API to not expose indices/tables.  Consider:
- Some idiot imports a PexFile, and sets all indices to 1.  Writes it out.  Oh no, corrupt pex file.  The user is in charge of making each index unique, and it's corrupt when they don't.  Okay well they're just an idiot.. but what about:
- Someone imports a pexfile, removes a few DebugFunctions, and exports.  That's a corrupt (or bloated?) pexfile, too.  Currently, the burden is on them to ALSO check what indices were associated with the DebugFunction classes they removed, and also remove them from the StringsTable.
- Opposite is true, too.  If someone adds a few DebugFunctions, they're in charge of "syncing" the strings table to have the appropriate entries.

The exposure of indices/StringsTable offers no interesting additional power to the user, really.  **It's just some homework they can fuck up.**  I sort of see the concept as a whole as derivative, aka, reproducible from the other data at hand.  We don't need to store or require the user to update strings tables.  We can construct them ourselves at the time of export just fine -> They're derivative data.

Ideally, we just expose string members, and hide away the on-disk detail that indices + table even exist.   If the user removes two DebugFunctions and exports, that's totally cool, because we're going to create a new string table at the time of export containing only the strings present in the PexFile object at the time of export.  No chance for desync, and real simple API for the user.  

## PexParsingException
I do like custom exceptions, but only when they include additional pieces of data.  RecordException includes the FormKey where the problem occurred, etc.   If it's not adding anything new, I like just using an existing exception that makes sense, so there's not just hundreds+ of shell exceptions for every possible error in existence.  `DataMisalignedException` or `InvalidDataException` would work just fine, for ex.    Alternatively, we could go the other way instead, and enrich PexParsingException a bit to include that extra data to make it useful as a distinct idea.  VariableType + Argument?  Not really sure what's relevant, but stuff like that.  

## New Projects
I think they can just be absorbed.  Pex stuff isn't importing anything at all so pretty clean for it to just be a namespace within `Mutagen.Bethesda.Core`.  Tests can just hop in the existing unit test project, maybe with their own subfolder for all the goods.

PR #195, opened 2021-04-01
Add comments identifying which aspects a field can be accessed by.

mrudat:


Mutagen.Bethesda.Generation/Modules/Aspects/AspectInterfaceDefinition.cs
```diff
@@ -12,6 +12,8 @@ public class AspectInterfaceDefinition
         public Func<ObjectGeneration, IEnumerable<(LoquiInterfaceDefinitionType Type, string Interface)>> Interfaces;
         public List<(LoquiInterfaceType Type, string Name, Action<ObjectGeneration, FileGeneration> Actions)> FieldActions = new List<(LoquiInterfaceType Type, string Name, Action<ObjectGeneration, FileGeneration> Actions)>();
         public Func<ObjectGeneration, bool> Test { get; set; }
+        public Func<ObjectGeneration, IEnumerable<TypeGeneration>> IdentifyFields { get; set; }
```
Noggog: I think this logic could have the default logic of looping over the `FieldActions` member, and finding TypeGenerations that match that list's `Name` field.  This would cover most of the default cases, and you could probably delete all the existing logic you're assigning to IdentifyFields in the aspect definitions.   This field would only be set to something else in very odd scenarios

PR #245, opened 2021-10-10
`Location.ActorCellUniques` and `Location.LocationCellUniques` should be LocationCellUnique

### LINKED ISSUE #244: `Location.ActorCellUniques` and `Location.LocationCellUniqes` have the wrong type
I think their type is meant to be `LocationCellUnique` rather than `LocationReference`. As things stand, copying a location that has unique references will cause all of them to lose their locations.

OddDrifter:
`Location.ActorCellUniques` and `Location.LocationCellUniques` are not LocationReferences. Trying to copy a location that has either of those fields causes the references in them to their locations. Should close #244. 

Screenshot 2021-10-10 140209

### DIFF
--- a/Mutagen.Bethesda.Skyrim/Records/Major Records/Location.xml
+++ b/Mutagen.Bethesda.Skyrim/Records/Major Records/Location.xml
@@ -9,8 +9,8 @@
           <Interface>IPlacedSimple</Interface>
         </FormLink>
       </List>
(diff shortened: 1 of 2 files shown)


Noggog:
Thanks for the fix!  Been on vacation so just got the chance to look at it.
I dont think ive associated your name over on discord, but feel free to swing by!

PR #362, opened 2022-08-21
Implemented VoiceTypeAssetLookup and moved it to Mutagen.Bethesda.Skyrim.Records.Assets

Elscrux:


Mutagen.Bethesda.UnitTests/Skyrim/Assets/VoiceTypeAssetLookupTestSkyrim.cs
```diff
@@ -0,0 +1,122 @@
...
+using Mutagen.Bethesda.Plugins;
+using Mutagen.Bethesda.Plugins.Cache;
+using Mutagen.Bethesda.Skyrim;
+using Mutagen.Bethesda.Skyrim.Records.Assets.VoiceType;
+using Xunit;
+namespace Mutagen.Bethesda.UnitTests.Skyrim.Assets;
+
+public class VoiceTypeAssetLookupTestSkyrim
+{
+    private static readonly ILinkCache<ISkyrimMod, ISkyrimModGetter> LinkCache = GameEnvironment.Typical.Skyrim(SkyrimRelease.SkyrimSE).LinkCache;
```
Noggog, line 12: Super happy to see some unit tests!

Unfortunately, I don't think using GameEnvironment will be a good route to use.   It'll work on local setups like yours or mine, but will fail once pushed up to GitHub and run on the CI pipeline where there's no skyrim installation to find, lol.

Definitely would be easier if we were running on a setup that had the game installed, but that probably won't ever happen.

I think instead, my typical two strategies are:
- Upload raw hex files that I made/copied by hand in xedit or whatever.  Those small test files can be committed and thus accessible by the server during test time.  This is good when we're testing gritty internals and malformed files that cannot be expressed by Mutagen classes (because they're corrupt bytes), etc.
- Just create the expected records in code.   Can make a new DialogTopic and DialogResponses with the situation to be tested just within the unit test itself and then test with them.  This seems to be the best route for this PR, id say

Mutagen.Bethesda.Skyrim/Records/Assets/VoiceType/VoiceTypeAssetLookup.cs
```diff
@@ -0,0 +1,906 @@
...
+
+
+    //Global variables
+    private IQuestGetter _currentQuest = new Quest(FormKey.Null, SkyrimRelease.SkyrimSE);
+    private ModKey _currentMod = ModKey.Null;
+
+    public void Prep(IAssetLinkCache linkCache)
+    {
+        _formLinkCache = linkCache.FormLinkCache;
+        _formLinkCache.Warmup<INpcGetter>();
```
Noggog: Don't think we need to Warmup here.    Warmup is mostly useful for when there's some distance between the Warmup event and when we used the results.  Say, a UI that knows it'll query Npcs for a FormKey picker, and so does a warmup on a background thread at startup to get those prepped, so the user's first query (which might be 5 minutes later) is instant.   But here in this backend code, with or without this Warmup we're going to query the npcs and do the work on this thread.

Mutagen.Bethesda.Skyrim/Records/Assets/VoiceType/VoiceTypeAssetLookup.cs
```diff
@@ -0,0 +1,906 @@
...
+    private readonly Dictionary<ModKey, VoiceContainer> _defaultSpeakerVoices = new();
+    private readonly Dictionary<ModKey, HashSet<string>> _defaultVoiceTypes = new();
+    private readonly Dictionary<FormKey, HashSet<string>> _speakerVoices = new();
+
+    private readonly Dictionary<FormKey, string> _voiceTypeEditorID = new();
+    private readonly Dictionary<FormKey, HashSet<FormKey>> _factionNPCs = new();
+    private readonly Dictionary<FormKey, HashSet<FormKey>> _classNPCs = new();
+    private readonly Dictionary<FormKey, HashSet<FormKey>> _raceNPCs = new();
+    private readonly Dictionary<bool, HashSet<FormKey>> _genderNPCs = new();
+    private HashSet<FormKey> _childNPCs = new();
```
Noggog: `_childNpcs` is probably a good place to use the null forgiveness operator.
`private HashSet<FormKey> _childNPCs = null!;`

We know we have a gentlemens agreement that Prep needs to be called before using this class (and this job is handled internally by our own library), and that we're setting _childNpcs within that, so any `new()` on this line is just wasted allocations we're going to immediately release to the GC

Alternatively, you could leave the `new()` and just have the `Build child cache` step add all the values to the hashset rather than making a new one

Mutagen.Bethesda.Skyrim/Records/Assets/VoiceType/VoiceTypeAssetLookup.cs
```diff
@@ -0,0 +1,906 @@
...
+public class VoiceTypeAssetLookup : IAssetCacheComponent
+{
+    private ILinkCache _formLinkCache = null!;
+
+    //Databases
+    private readonly Dictionary<ModKey, VoiceContainer> _defaultSpeakerVoices = new();
+    private readonly Dictionary<ModKey, HashSet<string>> _defaultVoiceTypes = new();
+    private readonly Dictionary<FormKey, HashSet<string>> _speakerVoices = new();
+
+    private readonly Dictionary<FormKey, string> _voiceTypeEditorID = new();
```
Noggog: Don't need this dictionary, as link caches come with a TryResolveIdentifier which can map from FormKey <-> Edid

Mutagen.Bethesda.Skyrim/Records/Assets/VoiceType/VoiceTypeAssetLookup.cs
```diff
@@ -0,0 +1,906 @@
...
+        {
+            case Condition.Function.GetIsID:
+                if (!functionConditionData.ParameterOneRecord.FormKey.IsNull)
+                {
+                    voices = new VoiceContainer(functionConditionData.ParameterOneRecord.FormKey, _speakerVoices[functionConditionData.ParameterOneRecord.FormKey]);
+                }
+
+                break;
+            case Condition.Function.GetIsVoiceType:
+                if (_voiceTypeEditorID.ContainsKey(functionConditionData.ParameterOneRecord.FormKey))
```
Noggog: Two things:
1)  `TryGetValue` should be used as it'll be one query into the dictionary instead of two (Contains + Get).
2)  Don't need _voiceTypeEditorID dictionary.  Instead just use
```
if (_formLinkCache.TryResolveIdentifier<IVoiceTypeGetter>(functionConditionData.ParameterOneRecord.FormKey, out var edid))
{
    //One voice type
    voices = new VoiceContainer(edid!);
}
```
Although, just realized there's a bug in the API interface, so you'll need that `edid!` null forgiveness operator.  Made a card for that
https://github.com/Mutagen-Modding/Mutagen/issues/364

Mutagen.Bethesda.Skyrim/Records/Assets/VoiceType/VoiceTypeAssetLookup.cs
```diff
@@ -0,0 +1,906 @@
...
+            baseVoices = new VoiceContainer(_speakerVoices);
+        }
+
+        baseVoices.Remove(voiceContainer);
+        return baseVoices;
+    }
+
+    private HashSet<string> GetVoiceTypes(FormKey speaker)
+    {
+        return _speakerVoices.ContainsKey(speaker) ? _speakerVoices[speaker] : new HashSet<string>();
```
Noggog: We can probably improve this if we make the return type `IReadOnlyCollection<string>` or `IEnumerable<string>`

This way, we can just use `Array.Empty<string>()` instead of allocating a new empty hashset in the false logic branch

Of course, downstream will need to adjusted to take IReadOnlyCollection, too, but that should all align after that

Mutagen.Bethesda.Skyrim/Records/Assets/VoiceType/VoiceContainer.cs
```diff
@@ -0,0 +1,250 @@
...
+        var removeVoiceTypes = new HashSet<string>();
+
+        foreach (var (voiceType, npcs) in _voices)
+        {
+            if (other._voices.ContainsKey(voiceType))
+            {
+                if (npcs.Any())
+                {
+                    //We don't have all NPCs of this voice type
+                    if (other._voices[voiceType].Any())
```
Noggog: Hmm, might be fine as-is, depending on the known context and whether we are okay with assuming `other._voices[voiceType]` will exist.   If not, this would be more appropriate if the key can potentially be missing:
`if (other._voices.TryGetValue(voiceType, out var valsForVoiceType) && valsForVoiceType.Any())`

Mutagen.Bethesda.Skyrim/Records/Assets/VoiceType/VoiceContainer.cs
```diff
@@ -0,0 +1,250 @@
...
+        IsDefault = false;
+    }
+
+    public void Insert(VoiceContainer other)
+    {
+        if (IsDefault && other.IsDefault) return;
+
+        foreach (var (voiceType, npcs) in other._voices)
+        {
+            if (_voices.ContainsKey(voiceType))
```
Noggog: `TryGetValue` is preferred.   `ContainsKey` is generally rarely used id say, as it's not too often you don't also care about the value associated with that key

Mutagen.Bethesda.Skyrim/Records/Assets/VoiceType/VoiceTypeAssetLookup.cs
```diff
@@ -0,0 +1,906 @@
...
+    private readonly Dictionary<FormKey, int> _dialogueSceneAliasIndex = new();
+
+
+    //Caches
+    private readonly Dictionary<FormKey, VoiceContainer> _questCache = new();
+    private readonly Dictionary<FormKey, HashSet<FormKey>> _sharedInfosCache = new();
+
+
+    //Global variables
+    private IQuestGetter _currentQuest = new Quest(FormKey.Null, SkyrimRelease.SkyrimSE);
```
Noggog: I think I'd prefer these to be local variables that are passed into the interested functions via parameters.   Having these global variables can lead to subtle bugs if/when function calls are reordered by refactoring that seems safe.   By having them be function parameters, proper order of calls is somewhat enforced by the compiler.

Also, if we ever want to add a Parallel call to this logic to speed stuff up in the future, having these as local variables will be much more threadsafe and flow into that pattern much easier

Mutagen.Bethesda.Skyrim/Records/Assets/VoiceType/VoiceTypeAssetLookup.cs
```diff
@@ -0,0 +1,906 @@
...
+    private ILinkCache _formLinkCache = null!;
+
+    //Databases
+    private readonly Dictionary<ModKey, VoiceContainer> _defaultSpeakerVoices = new();
+    private readonly Dictionary<ModKey, HashSet<string>> _defaultVoiceTypes = new();
+    private readonly Dictionary<FormKey, HashSet<string>> _speakerVoices = new();
+
+    private readonly Dictionary<FormKey, string> _voiceTypeEditorID = new();
+    private readonly Dictionary<FormKey, HashSet<FormKey>> _factionNPCs = new();
+    private readonly Dictionary<FormKey, HashSet<FormKey>> _classNPCs = new();
```
Noggog, line 17: These might be nice to swap to `IFormLinkGetter<INpcGetter>` (or whatever types they're supposed to be).  Just help keep the strong typing around and help ensure we're not accidentally crossing wires and putting Class formkeys into Faction lists, etc.

Mutagen.Bethesda.Skyrim/Records/Assets/VoiceType/VoiceTypeAssetLookup.cs
```diff
@@ -0,0 +1,906 @@
...
+                    _currentQuest = currentQuest;
+                    userConditions.Insert(GetVoices(responseContext.Record.Conditions));
+                }
+
+                voices.Merge(userConditions);
+            }
+
+
+            var responseFormID = responses.FormKey.ID.ToString("X8");
+            if (voices.IsEmpty()) Console.WriteLine($"WARNING: {responseFormID} has no valid voice types {topic.EditorID}");
```
Noggog: Don't want a Console write here.  This is within the context of a library, and we have no idea what system the user will be using our logic within.   Maybe it's a UI that wants to display this warning somewhere.  Or if it is a CLI, they might not want it printing this message and have to do work to filter it out somehow.

Do we need to throw an error/warning here?  We're not changing the logic either way.  Might just be appropriate to do as an [Analyzer](https://github.com/Mutagen-Modding/Mutagen.Bethesda.Analyzers)

Mutagen.Bethesda.Skyrim/Records/Assets/VoiceType/VoiceTypeAssetLookup.cs
```diff
@@ -0,0 +1,906 @@
...
+
+            if (topic.Subtype == DialogTopic.SubtypeEnum.SharedInfo && _sharedInfosCache.ContainsKey(responses.FormKey))
+            {
+                var userConditions = new VoiceContainer(true);
+                foreach (var responseKey in _sharedInfosCache[responses.FormKey])
+                {
+                    var responseContext = _formLinkCache.ResolveSimpleContext<IDialogResponsesGetter>(responseKey);
+                    if (responseContext is not { Parent.Record: {} }) continue;
+
+                    var currentTopic = (IDialogTopicGetter) responseContext.Parent.Record;
```
Noggog: The cast here assumes its parent is a DialogTopic (which it very well may always be, lol).  But still, safer to maybe swap to do
`if (!responseContext.TryGetParent<IDialogTopicGetter>(out var currentTopic)) continue;`

Noggog (commented):
Looking good overall!  Just dumped a heavy truck of anal detail thoughts.  I don't mind doing the changes myself if that's preferable for you.

Will also want to integrate into the DialogTopic's GetResolvedAssetLinks call, which is currently not compiling and throwing an exception

Mutagen.Bethesda.Skyrim/Records/Assets/VoiceType/VoiceTypeAssetLookup.cs
```diff
@@ -0,0 +1,906 @@
...
+                        _factionNPCs[factionKey].Add(npc.FormKey);
+                    } else
+                    {
+                        _factionNPCs.Add(factionKey, new HashSet<FormKey> { npc.FormKey });
+                    }
+                }
+
+                foreach (var classKey in GetClasses(npc))
+                {
+                    if (_classNPCs.ContainsKey(classKey))
```
Noggog: These can be simplified a bit to:
```cs
_classNPCs
    .GetOrAdd(classKey) // Can add a lambda if the value type didn't have a parameterless ctor
    .Add(npc.FormKey);
```

Mutagen.Bethesda.UnitTests/Skyrim/Assets/VoiceTypeAssetLookupTestSkyrim.cs
Elscrux: This one is an example where you really need unit tests in development. With all these recursive calls, it is super easy to make a change that breaks something.
GitHub CI is an issue, but creating the records in code won't be the solution because the voice type lookup references so many other records in the background.
The simplest approach might be to store all records used in all tests, but as it needs to go through all npcs, all factions, and so on, it would almost be the easiest approach to use Skyrim.esm. Only if that doesn't create any issues with GitHub or the pipeline of course. When more test files would be needed that have a similar scope of records, it would also quickly reach and go beyond the size of importing Skyrim.esm once for use in unit tests.

Mutagen.Bethesda.Skyrim/Records/Assets/VoiceType/VoiceTypeAssetLookup.cs
Elscrux: That'd be nice for sure. I used raw form keys initially because I wanted to save performance where possible. But I'm not sure it would actually make a difference performance-wise, so I'll do that if think that should be fine.

Mutagen.Bethesda.Skyrim/Records/Assets/VoiceType/VoiceTypeAssetLookup.cs
Elscrux: That is mostly a leftover from testing so I'll just remove it.

Mutagen.Bethesda.Skyrim/Records/Assets/VoiceType/VoiceTypeAssetLookup.cs
Elscrux: Ah, good to know!

Mutagen.Bethesda.Skyrim/Records/Assets/VoiceType/VoiceTypeAssetLookup.cs
Noggog: Hmm, might save on a little GC if/when the cache is cleaned up (as FormLinks are objects), but probably not worth worrying about too much overall

Mutagen.Bethesda.UnitTests/Skyrim/Assets/VoiceTypeAssetLookupTestSkyrim.cs
Noggog: Definitely agree unit tests are desirable and we should find a way to have them.

Are you suggesting we add Skyrim.esm to the repository?  I don't think that'll be viable, not only for the large file concerns, but also just copyright problems.

We need to generate our own test data, I think.  We may use existing Skyrim records as inspiration, but I dont think we should use them directly.   This is sort of what I was thinking:
```cs
    [Theory]
    [MutagenAutoData]
    public void TestAliasAdditionalVoicesNPCList(
        ModKey master,
        uint aliasId)
    {
        var mod = new SkyrimMod(master, SkyrimRelease.SkyrimSE);
        var topic2 = mod.DialogTopics.AddNew();
        var scene = mod.Scenes.AddNew();
        scene.Actions.Add(new SceneAction()
        {
            Type = SceneAction.TypeEnum.Dialog,
            Topic = topic2.ToNullableLink<IDialogTopicGetter>(),
            ActorID = (int)aliasId,
        });
        var response2 = new DialogResponses(mod);
        topic2.Responses.Add(response2);
        var quest = mod.Quests.AddNew();
        var voiceType = mod.VoiceTypes.AddNew();
        voiceType.EditorID = "FemaleCommander";
        var npc = mod.Npcs.AddNew();
        npc.Voice.SetTo(voiceType);
        var voiceType2 = mod.VoiceTypes.AddNew();
        voiceType2.EditorID = "MaleBrute";
        var npc2 = mod.Npcs.AddNew();
        npc2.Voice.SetTo(voiceType2);
        var formList = mod.FormLists.AddNew();
        formList.Items.Add(npc);
        formList.Items.Add(npc2);
        var alias = new QuestAlias()
        {
            ID = aliasId
        };
        alias.VoiceTypes.SetTo(formList);
        quest.Aliases.Add(alias);
        topic2.Category = DialogTopic.CategoryEnum.Scene;
        topic2.Quest.SetTo(quest);
        
        var linkCache = mod.ToImmutableLinkCache();
        var sut = new VoiceTypeAssetLookup();
        sut.Prep(linkCache.CreateImmutableAssetLinkCache());
        
        Assert.Equal(
            new VoiceContainer(new HashSet<string>
            {
                "FemaleCommander",
                "MaleBrute"
            }),
            sut.GetVoicesWithQuest(topic2, response2)
        );
    }
```
This seems to pass along the same logic routes that the original did.

Now, this is definitely verbose and ugly af in the above form, but we can refactor and improve the systems a bit to be more digestible.  But the fundamental concept is that we recreate the setup we're testing via code instead of relying on Skyrim.esm.   VoiceTypeAssetLookup is pretty complex, so there's lots of edge cases and situations we'd want to construct and test, for sure.

I got a few ideas to maybe help simplify the above test a bit via AutoFixture improvements, and a few other things.  I'll tinker around

Noggog: Toying around with stuff locally a bit, and adding some AutoFixture features, got it to look more like:

```cs
    [Theory]
    [MutagenModAutoData]
    public void TestAliasAdditionalVoicesNPCList(
        SkyrimMod mod,
        Scene scene,
        DialogTopic topic,
        Quest quest,
        DialogResponses dialogResponses,
        VoiceType voiceType,
        VoiceType voiceType2,
        Npc npc,
        Npc npc2,
        FormList formList,
        string editorId1,
        string editorId2,
        uint aliasId)
    {
        topic.Responses.Add(dialogResponses);
        topic.Category = DialogTopic.CategoryEnum.Scene;
        topic.Quest.SetTo(quest);
        
        scene.Actions.Add(new SceneAction()
        {
            Type = SceneAction.TypeEnum.Dialog,
            Topic = topic.ToNullableLink<IDialogTopicGetter>(),
            ActorID = (int)aliasId,
        });
        
        voiceType.EditorID = editorId1;
        npc.Voice.SetTo(voiceType);
        
        voiceType2.EditorID = editorId2;
        npc2.Voice.SetTo(voiceType2);
        
        formList.Items.Add(npc);
        formList.Items.Add(npc2);
        
        var alias = new QuestAlias()
        {
            ID = aliasId
        };
        alias.VoiceTypes.SetTo(formList);
        quest.Aliases.Add(alias);
        
        var linkCache = mod.ToImmutableLinkCache();
        var sut = new VoiceTypeAssetLookup();
        sut.Prep(linkCache.CreateImmutableAssetLinkCache());
        
        Assert.Equal(
            new VoiceContainer(new HashSet<string>
            {
                editorId1,
                editorId2
            }),
            sut.GetVoicesWithQuest(topic, dialogResponses)
        );
    }
```
Essentially, the new tooling is instantiating and abstracting away all the busiwork of making these new records and adding them to the mod.  Just ask for a new record, and it'll add it to the mod object (if it's a top level record) and give it to you for access.

If there's a commonality between all the tests where they always want XYZ set up, we can push the idea even further and add customization for this specific suite of tests

Elscrux: I'm not sure this will scale very well. I also made unit tests for Beyond Skyrim Cyrodiil. Using a stripped version of that plugin will be easier to use than adding everything by hand I think. Seeing as you usually also have to account for assigning npcs to factions, classes, placing npcs and adding them as quest alias and much more.

Noggog: Let's move to discord, lol

PR #363, opened 2022-08-22
add missing BipedObjectFlag values

TokcDK:


Noggog:
Hey!   Thanks a ton for taking the time to set up a PR!

There's a few things I'd like to talk over about the gritty details before I merge it in

# Dealing with "Unknown" Enum Values
My existing pattern for dealing with potentially unknown enum values is to just omit them, which is a bit different than how xedit seems to handle it.   For example, just to pick a random example out of a hat, let's look at MusicType's flags:

In xEdit:
https://github.com/TES5Edit/TES5Edit/blob/2f3c3950322c354204ca7968e6f7d2d4d92f3dcd/Core/wbDefinitionsTES5.pas#L9831

In Mutagen:
https://github.com/Mutagen-Modding/Mutagen/blob/dev/Mutagen.Bethesda.Skyrim/Records/Major%20Records/MusicType.cs

You'll notice xEdit defines an `Unknown 4` with value `0x10`, while Mutagen just opts to not list that entry at all.

Why the difference?   Well, you can still interact with `Unknown 4` in Mutagen, even if it's not listed.  Since all enums are `int`s, you can do stuff like:
```cs
if (someMusic.Flags.HasFlag((MusicType.Flag)0x10))
{
   // Found the oddball flag that no one knows what it is
}
```
And since an interested patcher can access those oddball values if they're truly interested, still, I opted to just omit them and not "clog up" the enum definitions with unknown values.  Keep them trim and just list the values the community knows a meaningful name for.

# BipedObjectFlag
So, with the above thoughts in mind, we need to decide if these new flags are "known" values or "unknown" values in the sense that they should be defined.  

Eg, is this:
```cs
if (someObj.Flags.HasFlag(BipedObjectFlag.BodyAddOn5))
{
}
```
More useful and informative to the coder than
```cs
if (someObj.Flags.HasFlag((BipedObjectFlag)0x000_10000))
{
}
```

My gut says that the coder will not know without some research what "slot" BodyAddOn5 actually associates with.  They'd have to look up and find out "oh, it's 0x000_10000, three slots after Ears.  Great, that's what I wanted"

An enum value of `Ears` is important because everyone knows what ears are and it's nice and easy to just have that label.   But `BodyAddon5` and `0x000_10000` are just random numbered slots that are probably used for miscellaneous different things.

But, I also am super detached from actual usage sometimes.  People dealing with bodyslots might appreciate the extra values in a way that isn't apparent just glancing at it for a sec like I am.   If you got some example code where it would be useful to have, definitely shoot a link and we can go over some specifics

TokcDK:
I have a lot of armor mods using this extra slots like splitted
I have modified version of Lootification patcher which I changed to work with additional slots and distribute armor sets usings this armors. 
But java patchers is very slow and because after I found Synthesis, after some time I started to make implementation of Lootification mod in C# because of much faster work but encountered with missing extra slots.

About info, I used info about slots from [here](http://wiki.tesnexus.com/index.php/Skyrim_bodyparts_number) and [here](https://www.loverslab.com/blogs/entry/14631-the-basics-of-body-slots-for-skyrim/) and changed some armor parts to use slots by the info.
What I see in my game
60 - just some body addons or accessories
59 - leg or forearm addons
58 - same as with 59, arm or leg addon
57 - see body addon part of armor set
56 - arm accessory, belt
55 - some face addon
54 - some arm accessory or face masks
53 - usually pants(outwear)
52 - usually pants(underwear)
51 - rare some accessory
50 - rare some accessory
49 - usually skirts here
48 - some misc
47 - backpacks and kind of
46 - scarfs and cloaks(40,46)
Mostly it equals extra slots infos.

Slot naming I suppose I got from [skyproc](https://github.com/tstavrianos/skyproc/blob/master/src/skyproc/genenums/FirstPersonFlags.java).

Maybe it will be usefull to add /// info for extra slots to be displayed in popup info.

Noggog:
Cool, so it sounds like you do use the extra slots.   But I'd say the naming is still unclear and a bit arbitrary for these extra slots, which makes me feel like we should omit the names for these?  Both of the documentation links you showed don't have entries for these unknown middle values, and the official slots as defined by Bethesda are all defined in Mutagen.

> after some time I started to make implementation of Lootification mod in C# because of much faster work but encountered with missing extra slots.

Does the code solution I mentioned initially of using the numeric values of the "unknown" slots you want to access work in your situation?  I feel like it should be workable?  (Probably just need to add a wiki entry about the topic, perhaps)

TokcDK:
> Cool, so it sounds like you do use the extra slots. But I'd say the naming is still unclear and a bit arbitrary for these extra slots, which makes me feel like we should omit the names for these? Both of the documentation links you showed don't have entries for these unknown middle values, and the official slots as defined by Bethesda are all defined in Mutagen.

Yes, as I mentioned, I use many armor packs with this extra slots and in game I see many items from different mods using this extra slots and because of it I extended Lootification code to use extra slots to make possible for the armors to be distributed. 
I did say the BodyAddon with /// popup with description would be useful. Documentation contains overall info to mod authors not to set slots randomly but using the infos and set it with more sense like case with slot 44 for mouth mask described with examples on LL.



> Does the code solution I mentioned initially of using the numeric values of the "unknown" slots you want to access work in your situation? I feel like it should be workable? (Probably just need to add a wiki entry about the topic, perhaps)

I created copy of the biped object flags like in this commit and using casting when it need because use digitals is very unconvenient because need to search this search and search slots info to use it with right way. It would be very useful if this slots was avalaible and had info about usual using. When some functions in mutagen have not descriptions it sometime is very hard to understand what they are doing but in this case we have mssing flags and missing descriptions and mod authors have a big chance to just drop to try to search all the info about slots.
Even if exist some oficial slots using for standart HCGBS armors other slots are also using by modders very often for a long time and it would be very useful to have it in Mutagen api for patch mod creation atleast ones which can be using in mods.
But if use bodyaddon names then need some short /// descriptions for the slots.

Noggog:
> like case with slot 44 for mouth mask described with examples on LL.
> It would be very useful if this slots was avalaible and had info about usual using

So, just to be clear, is slot 44 being for `MouthMask` either:
1)  A convention that LL and some of the modding community has set up that slot 44 is usually used for `MouthMask`s when using mods found on LL.
2)  Hardcoded by the engine somehow that 44 is meant for `MouthMask`s, and nothing else.

If it's 1), Mutagen really does not want to have an opinion that.  I dont want to define slot 44 to be `MouthMask` when it could be used by another modder just as well for `ElbowPad` in their setup. Know what I mean?

If you want comments to be laid out about all the various conventions modders might come up with for how to use these extra slots, I'd rather just have the comment link to a [Discussion](https://github.com/Mutagen-Modding/Mutagen/discussions) or something similar where people can post and discuss the various conventions slots may or may not be used for in certain contexts.  "For LL style mods, these slots are related to XYZ.  For OtherStyle mods, these slots are related to ABC.  +5 other conventions"

Noggog:
Hey TokcDK!

I'm going to close this out until we can do the research/discussion i mentioned above a bit more.  Feel free to continue commenting here, or swing by the Discord to have a more direct convo about it all!

TokcDK:
I still think that BodyAddon naming variant as in Skyproc with added /// descriptions for them is most convenient variant because of no specific using of each slot but to know that is usually using in each specific slot will be always useful.
And I just though it can be useful for patchers authors to not spend many time for slot values search and more easy to convert Skyproc patchers for Synthesis.
In my Lootification implementation I just added full version of flags and using it instead of mutagen variant with casting into mutagen when it need.

PR #408, opened 2023-01-21
fix namespaces

Jampi0n:
I tried the pre-release and noticed that I had to include additional namespace to use the new features, because I kept the ide-generated namespaces that just follow the folder structure.

For the files that are supposed to be used by the user, I changed the namespace to  Mutagen.Bethesda.Skyrim according to the other similar files.
For the file that is not supposed to be used by the user (GetContainingCell.cs) I used  Mutagen.Bethesda.Skyrim.Internals instead.

### DIFF
--- a/Mutagen.Bethesda.Skyrim/Extensions/PlacedContextExt.cs
+++ b/Mutagen.Bethesda.Skyrim/Extensions/PlacedContextExt.cs
@@ -1,6 +1,6 @@
 using System.Diagnostics.CodeAnalysis;
 using Mutagen.Bethesda.Plugins.Cache;
-using Mutagen.Bethesda.Skyrim.Internals;
+using Mutagen.Bethesda.Skyrim.Records.Tooling;
 using PlacedContext =
     Mutagen.Bethesda.Plugins.Cache.IModContext<Mutagen.Bethesda.Skyrim.ISkyrimMod,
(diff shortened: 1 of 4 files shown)


Noggog (changes requested):
Yeah, namespace choices are always hard.   Lots of subjective choices to be had.   Originally I had everything in the top `Mutagen.Bethesda.Skyrim` namespace, but that got unmanageable, so I started splitting things up.   

I left the base records themselves in the `Mutagen.Bethesda.Skyrim` namespace (as those were most used).  But this maybe is a mistake, and they should be in the `Mutagen.Bethesda.Skyrim.Records` subnamespace.  Eh, no real right or wrong, and fine enough as is, I guess.

But anyway, I do think it's fine to keep some of the more obscure other things besides the actual record classes in other namespaces.   So I don't mind `WorldspaceCellLocationCache` living in the `Records/Tooling` subnamespace.  (GetContainingCell is not consequential, as it's `internal` so no one will see it either way)

The only one I might change is the PlacedContextExt being in the `Mutagen.Bethesda.Skyrim` namespace so it "hooks in" cleanly.  Where once someone wants to use it, and see they need `WorldspaceCellLocationCache`, then the IDE will pull in the other namespace.

So maybe just leave the extension namespace tweak and revert the others?

PR #419, opened 2023-02-17
Fix invalid FormLink types

### LINKED ISSUE #418: Skyrim IOwner is implemented by PlacedNpc, not Npc

Elscrux:
IOwner interface was implemented by PlacedNpc instead of Npc #418

### DIFF
--- a/Mutagen.Bethesda.Skyrim/Documentation/LinkInterfaceDocumentation_Generated.md
+++ b/Mutagen.Bethesda.Skyrim/Documentation/LinkInterfaceDocumentation_Generated.md
@@ -177,7 +177,7 @@ A `FormLink<IItem>` could then point to all those record types by pointing to th
 - LeveledItem
 ### IOwner
 - Faction
-- PlacedNpc
+- Npc
 ### IPlaceableObject
 - AcousticSpace
 - Activator
@@ -467,6 +467,7 @@ A `FormLink<IItem>` could then point to all those record types by pointing to th
 - INpcOrList
 - INpcSpawn
 - IObjectId
+- IOwner
 ### ObjectEffect
 - IEffectRecord
 - IExplodeSpawn
@@ -476,7 +477,6 @@ A `FormLink<IItem>` could then point to all those record types by pointing to th
 - IKeywordLinkedReference
 - ILinkedReference
 - ILocationTargetable
-- IOwner
 - IPlaced
 - IPlacedSimple
 ### PlacedObject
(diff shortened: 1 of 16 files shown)


Elscrux:
Added fix for #420

Noggog:
# IOwner
This seems to have existed from before the condition granularity project.  Namely within Npc's XOWN record
https://github.com/TES5Edit/TES5Edit/blob/dev-4.1.5/Core/wbDefinitionsTES5.pas#L5899

In that, it lists Factions, Npcs, and PlacedNpcs as valid.  Not sure why Npcs weren't marked in Mutagen previously.. perhaps that line got updated after Skyrim definitions were defined

So, at the very least, Npc should be added as an IOwner

But then we think about IsOwner's condition function:
https://ck.uesp.net/wiki/IsOwner
Seems to only list Npc and Faction.

So.. is PlacedNpc actually valid for IOwner when in the context of an Npc's XOWN record?   If so, we need two link interfaces.  If not, we can see that as a bug in xedit and this PR's changes are good for IOwner.

Elscrux:
> # IOwner
> 
> This seems to have existed from before the condition granularity project. Namely within Npc's XOWN record https://github.com/TES5Edit/TES5Edit/blob/dev-4.1.5/Core/wbDefinitionsTES5.pas#L5899
> 
> In that, it lists Factions, Npcs, and PlacedNpcs as valid. Not sure why Npcs weren't marked in Mutagen previously.. perhaps that line got updated after Skyrim definitions were defined
> 
> So, at the very least, Npc should be added as an IOwner
> 
> But then we think about IsOwner's condition function: https://ck.uesp.net/wiki/IsOwner Seems to only list Npc and Faction.
> 
> So.. is PlacedNpc actually valid for IOwner when in the context of an Npc's XOWN record? If so, we need two link interfaces. If not, we can see that as a bug in xedit and this PR's changes are good for IOwner.

In the CK, the IsOwner condition can't be filled with a PlacedNpc or ReferenceAlias. That also wouldn't make a lot of sense, because there is no way to set the owner of a reference to a PlacedNpc. You can only set it to Npc or Faction.

Reference owner:
image

Cell owner:
image

Here are some links to the uesp file format pages that only mention faction and npc.
https://en.uesp.net/wiki/Skyrim_Mod:Mod_File_Format/ACHR
https://en.uesp.net/wiki/Skyrim_Mod:Mod_File_Format/CELL
https://en.uesp.net/wiki/Skyrim_Mod:Mod_File_Format/REFR

Elscrux:
I also ran some tests and didn't find any instances of PlacedNpcs being used as Owners in Skyrim's plugins.

Noggog (approved):

PR #423, opened 2023-03-06
Add absolute file path detection

Elscrux:
This part will help detecting string like those: `C:\Skyrim\Data\Meshes\Clutter\MyMesh.nif` and turn them into `Meshes\Clutter\MyMesh.nif`.

### DIFF
--- a/Mutagen.Bethesda.Core/Plugins/Assets/AssetLink.cs
+++ b/Mutagen.Bethesda.Core/Plugins/Assets/AssetLink.cs
@@ -39,39 +39,41 @@ public class AssetLinkGetter<TAssetType> : IComparable<AssetLinkGetter<TAssetTyp
 
     public string RawPath => _rawPath;
 
-    public string DataRelativePath
-    {
-        get {
-            var path = RawPath;
-            
-            // Reduce all absolute paths to the path under data directory
-            if (path.Contains(Path.VolumeSeparatorChar)) {
-                var dataDirectoryIndex = path.IndexOf(DataInfix, IAssetLinkGetter.PathComparison);
-                if (dataDirectoryIndex != -1) {
-                    path = path[(dataDirectoryIndex + DataInfix.Length)..];
-                } else {
-                    dataDirectoryIndex = path.IndexOf(DataInfixAlt, IAssetLinkGetter.PathComparison);
-                    if (dataDirectoryIndex != -1) {
-                        path = path[(dataDirectoryIndex + DataInfixAlt.Length)..];
-                    }
-                }
-            }
-            
-            path = path
-                .TrimStart(Path.DirectorySeparatorChar)
-                .TrimStart(Path.AltDirectorySeparatorChar);
-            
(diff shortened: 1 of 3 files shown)


Mutagen.Bethesda.Core/Plugins/Assets/AssetLink.cs
```diff
@@ -39,17 +41,31 @@ public AssetLinkGetter(string rawPath)
 
     public string DataRelativePath
     {
-        get
-        {
-            var trimmedPath = RawPath
+        get {
```
Noggog: If you can run this code through the code formatter that'd be good.  For the repos here, id like to keep it to the C# styling, so future devs don't get too crazy of diffs when they do any auto-formatting

Mutagen.Bethesda.Core/Plugins/Assets/AssetLink.cs
```diff
@@ -39,17 +41,31 @@ public AssetLinkGetter(string rawPath)
...
+            var path = RawPath;
+            
+            // Reduce all absolute paths to the path under data directory
+            if (path.Contains(Path.VolumeSeparatorChar)) {
```
Noggog: Maybe refactor out to a private function or two, for clarity?  Getting pretty complex for a property definition, imo.

Mutagen.Bethesda.Core/Plugins/Assets/IAssetLink.cs
```diff
@@ -14,11 +14,16 @@ public interface IAssetLinkGetter
...
     /// <summary>
-    /// Raw path relative to the game's data directory, as it is commonly known in the Skyrim modding community.
+    /// Raw path relative to the game's data directory, as it is commonly known in the Bethesda modding community.
+    /// <example>
+    /// <code>
     /// Example:
-    ///     \Data\Meshes\Clutter\MyMesh.nif => Meshes\Clutter\MyMesh.nif
-    ///     Data\Meshes\Clutter\MyMesh.nif  => Meshes\Clutter\MyMesh.nif
-    ///     Clutter\MyMesh.nif              => Meshes\Clutter\MyMesh.nif
+    ///     C:\Skyrim\Data\Meshes\Clutter\MyMesh.nif => Meshes\Clutter\MyMesh.nif
```
Noggog, line 21: I think this DataRelativePath is complex enough and has arbitrary and important enough logic to warrant unit tests.   I thought we'd have had some in the AssetLinkTests class, but not seeing any specifically related to testing DataRelativePath too much.

I think if we can get at least 4 tests, one for each standardization style listed here, that'd be enough?

Noggog (changes requested):

Mutagen.Bethesda.Core/Plugins/Assets/AssetLink.cs
Elscrux: oops, yeah I missed that

PR #432, opened 2023-04-21
Fix typo in Aggression enum.

NoIDontCare:


Noggog:
Thanks so much for the fix!

You can mix in the filename typo fix as you mentioned on discord.  Won't be a problem

NoIDontCare:
Done. Also fixed another typo I found in the enum. I don't see any other issues inside the enum.

PR #439, opened 2023-05-06
Asset API - Move static fields into non-generic interface

Elscrux:


Noggog:
I like the idea.  I think it's atypical to have these fields in an interface.  Not sure why or what, but just sets my C# convention spider senses tingling.   I think I'd rather just declare a `static class AssetLink` without a generic to house the fields, so people can do `AssetLink.TheStaticField`

PR #442, opened 2023-05-07
Add static abstract Instance property to IAssetType

Elscrux:


Mutagen.Bethesda.Core/Assets/IAssetType.cs
```diff
@@ -24,8 +28,30 @@ public interface IAssetType
...
+#if NET7_0
+        switch (gameRelease) {
+            case GameRelease.Oblivion:
+                break;
+            case GameRelease.SkyrimLE:
+            case GameRelease.SkyrimSE:
+            case GameRelease.SkyrimSEGog:
+            case GameRelease.SkyrimVR:
+            case GameRelease.EnderalLE:
+            case GameRelease.EnderalSE:
```
Noggog, line 41: This looks still WIP?  Probably fine, but just double checking

Mutagen.Bethesda.Core.UnitTests/Plugins/Assets/TestAssetType.cs
```diff
@@ -4,7 +4,11 @@ namespace Mutagen.Bethesda.UnitTests.Plugins.Assets;
 
 public class TestAssetType : IAssetType
 {
+#if NET7_0
```
Noggog: Let's flip this around, i think.
```
#if NET6_0
// old way
#else
// new way
#endif
```
That way when we upgrade to Net8, it'll still use the newer system

Mutagen.Bethesda.Core/Assets/IAssetType.cs
```diff
@@ -24,8 +28,30 @@ public interface IAssetType
     /// <param name="gameRelease">Release of the game this asset comes from</param>
     /// <param name="path">Path of the asset</param>
     /// <returns>Instance of the parsed asset type or null if no asset type could be determined</returns>
-    public static IAssetType? GetAssetType(GameRelease gameRelease, string path)
-    {
+    public static IAssetType? GetAssetType(GameRelease gameRelease, string path) {
```
Noggog, line 31: Not a huge deal, but would be good to run files through the IDE formatter for Mutagen-proper.    Just good to keep it on more official C# styling standards.  Otherwise there ends up being lots of unnecessary changesets down the line when someone finally runs the formatter

Noggog (changes requested):

PR #452, opened 2023-07-17
Add GetAssetType()

Elscrux:
Add GetAssetType() based on file extension and base folder. Tests are included.

@Noggog I had to include a limited version of ConvertToDataRelativePath(), without the base folder check as that already needs information about the asset type. I don't know how to properly do this without copy paste as I'm not very familiar with Spans. If you see a way to do that, feel free to change it :)

### DIFF
--- a/Mutagen.Bethesda.Core/Assets/IAssetType.cs
+++ b/Mutagen.Bethesda.Core/Assets/IAssetType.cs
@@ -1,6 +1,6 @@
-﻿using System;
-using System.Collections.Generic;
-
+﻿using System.Reflection;
+using Mutagen.Bethesda.Plugins.Assets;
+using Noggog;
 namespace Mutagen.Bethesda.Assets;
 
 /// <summary>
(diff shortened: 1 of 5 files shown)


Mutagen.Bethesda.UnitTests/Skyrim/Assets/AssetTypeTests.cs
```diff
@@ -41,4 +42,38 @@ public void TestAllImplementationsHaveNoBaseClass()
         }
     }
 #endif
+
+    [Fact]
+    public void TestGetAssetType()
+    {
+        IAssetType.GetAssetType(GameCategory.Skyrim, "Music\\Special\\Dungeon\\DungeonBoss01.xwm").Should().BeOfType<SkyrimMusicAssetType>();
```
Noggog: Linux bois are going to want `Path.Combine` usage here

Mutagen.Bethesda.Core/Assets/IAssetType.cs
```diff
@@ -22,36 +22,68 @@ public interface IAssetType
     /// </summary>
     IEnumerable<string> FileExtensions { get; }
 
+    private static readonly Dictionary<GameCategory, Dictionary<string,Dictionary<string,IAssetType>>> Types;
+
+    static IAssetType()
```
Noggog: Let's move this to another internal class's static constructor.  `GetAssetTypeLocator` or something else.

Putting it here in `IAssetType` means that this reflection code will be less lazy than it could be.  Some code might not be interested in calling `GetAssetType`, but still have to pay this startup price.   If we move it elsewhere, then it's only when `GetAssetType` gets called and accesses `GetAssetTypeLocator` that the reflection will be run

Mutagen.Bethesda.Core/Assets/IAssetType.cs
```diff
@@ -22,36 +22,68 @@ public interface IAssetType
...
-            case GameRelease.SkyrimVR:
-            case GameRelease.EnderalLE:
-            case GameRelease.EnderalSE:
-                
-                break;
-            case GameRelease.Fallout4:
-                break;
-            default:
-                throw new ArgumentOutOfRangeException(nameof(gameRelease), gameRelease, null);
+    public static IAssetType? GetAssetType(GameCategory gameCategory, string path)
```
Noggog: Let's rename this to TryGetAssetType.   The fact that it returns nullable means it's a TryGet, as far as Mutagen's patterns elsewhere are concerned.    A function named `GetAssetType` would be one that didn't return null, and threw an exception if it couldn't be converted.   Up to you if you want to define the variants

Noggog (changes requested):
Unit tests are failing.  Not immediately apparent why.   Just says that the GetAssetType expected non-null, but got null.. which isn't too helpful.

Maybe as a way to help dig into it, we can define some new unit tests that just flex, say:
- `ConvertToDataRelativePath`.  Maybe the issue is in there, and a more dedicated unit test will expose that
- Make the Types dictionary internal, and write a unit test that queries the dictionary directly to see if what we expect is in there

Hopefully one of those tests will shed light on why the current one is failing on the CI server

Elscrux:
I included the feedback about the code. Maybe the tests also work now because I use Path.Combine instead? I'll wait for the pipeline to run through and try debugging the tests further when they fail.

Elscrux:
@Noggog Tests should be fixed now. I limited everything to dotnet 7, as the reflection code used to build the dictionary specifically uses dotnet 7 code. There could also be support for dotnet 6 with some tweaks but I left it as it is for now.

If you have any more suggestions, or ideas how to reduce code duplication with the two ConvertToDataRelativePath() functions, let me know. Otherwise I think this is ready to be merged.

Noggog (approved):

PR #455, opened 2023-07-21
AssetLink remap improvements

Elscrux:


Mutagen.Bethesda.Core/Plugins/Assets/AssetLink.cs
```diff
@@ -251,6 +251,20 @@ public override string ToString()
         return DataRelativePath;
     }
 
+    public override int GetHashCode()
+    {
+        return AssetLink.PathComparer.GetHashCode(RawPath);
```
Noggog: Might want to add `typeof(TAssetType).GetHashCode()` to the mix.
There's a C# class called `HashCode` nowadays that helps build them with multiple items easier

Mutagen.Bethesda.Core/Plugins/Assets/IAssetLinkContainer.cs
```diff
@@ -28,6 +28,32 @@ public interface IAssetLinkContainer : IAssetLinkContainerGetter
...
+        IReadOnlyDictionary<IAssetLinkGetter, string> mapping,
+        IAssetLinkCache linkCache)
+    {
+        assetLinkContainerGetter.RemapAssetLinks(mapping, AssetLinkQuery.Resolved, linkCache);
+    }
+    
+    public static void RemapAllAssetLinks(
+        this IAssetLinkContainer assetLinkContainerGetter,
+        IReadOnlyDictionary<IAssetLinkGetter, string> mapping,
+        IAssetLinkCache? linkCache = null)
```
Noggog: I'd make this required here

Mutagen.Bethesda.Skyrim/Records/SkyrimMod.cs
```diff
@@ -30,7 +30,7 @@ partial class SkyrimModSetterCommon
         IReadOnlyDictionary<IAssetLinkGetter, string> mapping,
         AssetLinkQuery queryCategories)
     {
-        throw new NotImplementedException();
+        // Nothing to do here, we can't change the name of the mod
```
Noggog, line 33: Maybe throwing an exception if we detect a remap is requested might be more appropriate?  If we detect a match and it's being requested for a remap, then the user is asking for us to do something that isn't possible and doesn't make sense.   Skipping silently makes sense if they might ask that of us without realizing it, but putting in a remap for this resolved asset is something explicit enough to warrant a throw, I think.

Maybe we can define a new `AssetLinkResolvedRemapException` or something to the `Plugins/Exceptions` folder

Or maybe no one cares enough, lol.  I'm fine with that, too

Noggog (changes requested):

Mutagen.Bethesda.Skyrim/Records/SkyrimMod.cs
Elscrux: I don't think throwing is the right way to do it as you're going to end up calling RemapAllAssetLinks on a random record and you shouldn't get an exception like that. And currently, it's not even checking if any mapping exists or if we want to query for resolved asset links.

Mutagen.Bethesda.Core/Plugins/Assets/IAssetLinkContainer.cs
Elscrux: Ah yes, forgot to change that

Mutagen.Bethesda.Core/Plugins/Assets/AssetLink.cs
Elscrux: Ok makes sense

Mutagen.Bethesda.Skyrim/Records/SkyrimMod.cs
Noggog: Well the idea would be you'd only throw if there was a match in the `mapping` that was requesting the resolved asset be remapped.   If the user just calls the function normally it would flow through fine.. it would only be if they added a resolved asset to the mapping that it would throw a "can't do that"

But probably too much work for what it's worth, haha

Elscrux: Yeah, if there is a valid mapping that would make sense, but yeah I think it might not be worth adding the extra steps, just to throw an exception. Especially seeing as on most occasions the code adds extra steps for checking for the mapping when you're not actually interested in this.

Noggog: Yeah, we can revisit if someone wants it i suppose

Noggog (approved):

PR #464, opened 2023-08-20
ActorValueInformation -> ActorValue mapping

### LINKED ISSUE #447: ActorValueInformation -> ActorValue mapping
Add some utility mapping between the two concepts so you can get the enum if you have the record

MarkKoz:
Closes #447

Implement a `ActorValueMapper` to map a `ActorValueInformation` record to the corresponding value in the `ActorValue`  enum.

I exported the AVIF records from SSEEdit using a slightly modifed version of the `QuickDisplay` script from [here](https://www.nexusmods.com/skyrim/mods/49373). Then I imported the CSV into [this](https://gist.github.com/MarkKoz/c56ca64d31258803753e610945604dab) script to generate the code.


MarkKoz:
Some of these records are actually in `SkyrimSE.exe`. Does that mean the string given to the `FormKey` factory should use the exe instead of `Skyrim.esm`?

Noggog (approved):
Looking good!   

Can we do the same for Fo4 while we're here?

Also, might as well fix that typo you mentioned in the issue

Noggog:
> Some of these records are actually in SkyrimSE.exe. Does that mean the string given to the FormKey factory should use the exe instead of Skyrim.esm?

Mutagen doesn't actually know about the records within Skyrim.exe at all.  If it did, it probably would be hardcoded records defined by hand and included in `Mutagen.Bethesda.Skyrim` that were always resolvable?   But would need to do more research on how those records behave.  Can you override these records?  Any other restrictions?  #468 

Anyway, for now, I think the records within Skyrim.exe can be viewed as non-existent

MarkKoz:
I think Fallout 4 is better left for a separate issue/PR. The `ActorValue` enum itself is incorrect to begin with (it's just copied from Skyrim currently).

PR #465, opened 2023-08-20
generated_code for generated files in editorconfig

FlayaN:
Set generated_code = true for *_Generated.cs in .editorconfig

When doing my other PR i noticed that there were a lot of extra spaces in files that I manually omitted.

When running `dotnet format Mutagen.Records.sln` in the root folder it also formatted the generated files, this flag fixes that.

This flag should also stop accidental formatting of generated files both in VS and Rider.

So if you want this behaviour it would be nice if you ran the format command in dev. There are currently a couple of hundred files with extra spaces in the source. (I didn't want to push them in this PR due to maybe not being in sync with the code if and when this is merged.)

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -32,3 +32,6 @@ indent_style = space
 indent_size = 2
 trim_trailing_whitespace = true
 insert_final_newline = true
+
+[*_Generated.cs]
+generated_code = true


FlayaN:
Nevermind, seems to be an issue with using generated_code and nullable enable.. 
https://github.com/dotnet/roslyn/blob/main/docs/features/nullable-reference-types.md#generated-code
Closing this PR

Noggog:
Seems like it would be viable if we identify the files that break, and add the `#nullable restore`

seemed to fix it for this file:
image


> So if you want this behaviour it would be nice if you ran the format command in dev. There are currently a couple of hundred files with extra spaces in the source.

Yeah, I wouldnt mind getting all the files formatted.   I'd only want to worry about it though if we could get some "format on save" option to keep the incoming PRs formatted without hounding the users every time to do so.

PR #466, opened 2023-08-20
Add IKeyworded.HasKeywords extension methods for IEnumerables

### LINKED ISSUE #415: IKeyworded.HasKeywords
Right now it only offers convenience extension methods for single keyword checks.  Would be nice to be able to pass in a collection.

MarkKoz:
Close #415

We could make the non-list methods proxy to the list methods to reduce code duplication (by wrapping the single argument in a list), but wasn't sure it was worthwhile.

### DIFF
--- a/Mutagen.Bethesda.Core/Plugins/Aspects/IKeyworded.cs
+++ b/Mutagen.Bethesda.Core/Plugins/Aspects/IKeyworded.cs
@@ -287,8 +287,7 @@ namespace Mutagen.Bethesda
             }
             foreach (var keywordForm in keyworded.Keywords.Select(link => link.FormKey))
             {
-                TKeyword keyword;
-                if (cache.TryResolve(keywordForm, out keyword!))
+                if (cache.TryResolve<TKeyword>(keywordForm, out var keyword))
                 {
                     var kwEditorID = keyword.EditorID;
                     if (editorIDs.Any(editorID => kwEditorID?.Equals(editorID, stringComparison) ?? false))
--- a/Mutagen.Bethesda.Core/Plugins/Aspects/IKeyworded.cs
+++ b/Mutagen.Bethesda.Core/Plugins/Aspects/IKeyworded.cs
@@ -219,48 +219,12 @@ namespace Mutagen.Bethesda
         /// <param name="keyworded">Keyworded record to check</param>
         /// <param name="keywordKeys">FormKeys of the Keyword records to look for</param>
         /// <returns>True if the Keyworded record contains a Keyword link /w any of the given FormKeys</returns>
-        public static bool HasKeyword<TKeyword>(
+        public static bool HasAnyKeyword<TKeyword>(
             this IKeywordedGetter<TKeyword> keyworded,
             IEnumerable<FormKey> keywordKeys)
             where TKeyword : class, IKeywordCommonGetter
         {
             return keyworded.Keywords?.IntersectBy(keywordKeys, x => x.FormKey).Any() ?? false;
-        }
-        
-        /// <summary>
-        /// Checks if a Keyworded record contains any specific Keyword, by FormKey.
-        /// Also looks up that keyword in the given cache. <br />
-        /// <br />
-        /// Note that this function only succeeds if the record contains the keyword,
-        /// and the cache found it as well. <br />
-        /// <br />
-        /// It is possible that the record contains the keyword, but it could not be found
-        /// by the cache.
-        /// <br />
-        /// Aspects: IKeywordedGetter&lt;IKeywordCommonGetter&gt;
-        /// </summary>
-        /// <param name="keyworded">Keyworded record to check</param>
-        /// <param name="keywordKeys">FormKeys of the Keyword records to look for</param>
-        /// <param name="cache">LinkCache to resolve against</param>
-        /// <param name="keyword">Keyword record retrieved, if keyworded record does contain</param>
-        /// <returns>True if the Keyworded record contains a Keyword link /w any of the given FormKeys</returns>
-        public static bool TryResolveKeyword<TKeyword>(
-            this IKeywordedGetter<TKeyword> keyworded,
-            IEnumerable<FormKey> keywordKeys,
-            ILinkCache cache,
-            [MaybeNullWhen(false)] out TKeyword keyword)
-            where TKeyword : class, IKeywordCommonGetter
-        {
-            foreach (var keywordKey in keywordKeys)
-            {
-                if (cache.TryResolve(keywordKey, out keyword) && HasKeyword(keyworded, keywordKey))
-                {
-                    return true;
-                }
-            }
-
-            keyword = default;
-            return false;
         }
         
          /// <summary>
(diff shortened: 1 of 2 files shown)


Mutagen.Bethesda.Core/Plugins/Aspects/IKeyworded.cs
```diff
@@ -215,5 +210,185 @@ public static class IKeywordedExt
...
+        
+        /// <summary>
+        /// Checks if a Keyworded record contains any specific Keyword, by FormKey.
+        /// <br />
+        /// Aspects: IKeywordedGetter&lt;IKeywordCommonGetter&gt;
+        /// </summary>
+        /// <param name="keyworded">Keyworded record to check</param>
+        /// <param name="keywordKeys">FormKeys of the Keyword records to look for</param>
+        /// <returns>True if the Keyworded record contains a Keyword link /w any of the given FormKeys</returns>
+        public static bool HasKeyword<TKeyword>(
```
Noggog: Looking at this closer, I think renaming this to `HasAnyKeyword` will be a better name to describe its behavior.   Leaves open an alternative `HasAllKeywords` or somesuch if we want to add that later

Mutagen.Bethesda.Core/Plugins/Aspects/IKeyworded.cs
```diff
@@ -215,5 +210,185 @@ public static class IKeywordedExt
...
+        /// by the cache.
+        /// <br />
+        /// Aspects: IKeywordedGetter&lt;IKeywordCommonGetter&gt;
+        /// </summary>
+        /// <param name="keyworded">Keyworded record to check</param>
+        /// <param name="keywordKeys">FormKeys of the Keyword records to look for</param>
+        /// <param name="cache">LinkCache to resolve against</param>
+        /// <param name="keyword">Keyword record retrieved, if keyworded record does contain</param>
+        /// <returns>True if the Keyworded record contains a Keyword link /w any of the given FormKeys</returns>
+        public static bool TryResolveKeyword<TKeyword>(
```
Noggog: Maybe if we renamed this to `TryResolveAnyKeyword`, then that'd make more sense with its behavior?  Still a weird function to expose.   `HasAnyKeyword` is being added as it's so very often used, some people have requested it instead of writing the (still simple) Linq statement.

But `TryResolveAnyKeyword` is sort of escaping that reasoning, as it's not something I think people will be too interested in heavily using, and is sort of flooding the intellisense at that point.

At the bare minimum, it's used for the final HasAnyKeyword for editorIDs, so maybe it's useful as a private helper method?

Mutagen.Bethesda.UnitTests/Plugins/Aspects/IKeywordedTests.cs
```diff
@@ -124,4 +127,128 @@ public void TryResolveKeyword_ByEditorID_Found()
...
+        Npc npc = new Npc(TestConstants.Form1, SkyrimRelease.SkyrimSE);
+        Keyword keyword2 = new Keyword(TestConstants.Form2, SkyrimRelease.SkyrimSE);
+        Keyword keyword4 = new Keyword(TestConstants.Form4, SkyrimRelease.SkyrimSE);
+        npc.Keywords = new ExtendedList<IFormLinkGetter<IKeywordGetter>>();
+        npc.Keywords.Add(TestConstants.Form3);
+        npc.HasKeyword(ImmutableList.Create(keyword2, keyword4)).Should().BeFalse();
+    }
+
+    [Fact]
+    public void HasKeyword_ByRecordList_Found()
```
Noggog: Tests look good!  If you want to get fancy, feel free to try upgrading them to use the AutoFixture systems.  Let's you do stuff like this:
```
[Theory, MutagenModAutoData]
public void HasKeyword_ByRecordList_NotFound(Npc npc, Keyword keyword2, Keyword keyword3, Keyword keyword4)
{
    npc.Keywords = new ExtendedList<IFormLinkGetter<IKeywordGetter>>();
    npc.Keywords.Add(keyword3.ToLink());
    npc.HasKeyword(ImmutableList.Create(keyword2, keyword4)).Should().BeFalse();
}
```
Where the test systems make the objects for you and handle the uninteresting details that are unrelated to this test.  

The tech does lots of crazy stuff.  Haven't gotten around to documenting it all on the Mutagen wiki, yet

Noggog (changes requested):

Mutagen.Bethesda.Core/Plugins/Aspects/IKeyworded.cs
MarkKoz: I've renamed the methods and removed the `Try*` methods.

Mutagen.Bethesda.UnitTests/Plugins/Aspects/IKeywordedTests.cs
MarkKoz: Nice! I've updated the tests to use that. Wasn't sure how to make the most out of it for the editor ID tests.

Mutagen.Bethesda.Core/Plugins/Aspects/IKeyworded.cs
```diff
@@ -215,5 +210,95 @@ public static class IKeywordedExt
...
+            // ToDo
+            // Consider EDID link cache systems if/when available
+            if (keyworded.Keywords == null)
+            {
+                return false;
+            }
+            foreach (var keywordForm in keyworded.Keywords.Select(link => link.FormKey))
+            {
+                TKeyword keyword;
+                if (cache.TryResolve(keywordForm, out keyword!))
```
MarkKoz: Using the null forgiving operator to avoid a CS8600 error. Not sure if there is a better way to go about this. I cannot use `out TKeyword? keyword` since `TryResolve` does not accept a nullable version of the type.

Noggog: `if (cache.TryResolve(keywordForm, out var keyword))`
is the way I'd typically write it, as yeah, defining the variable in the line before makes things confused as it's sort of a quantum-null variable.   Perhaps changing L290 to be nullable might work?   But yeah, using `var` is probably the preferred route.

Mutagen.Bethesda.UnitTests/Plugins/Aspects/IKeywordedTests.cs
```diff
@@ -114,10 +103,9 @@ public void TryResolveKeyword_ByFormKey_Found()
         kw.Should().Be(keyword);
     }
 
-    [Fact]
-    public void TryResolveKeyword_ByEditorID_Found()
+    [Theory, MutagenModAutoData]
+    public void TryResolveKeyword_ByEditorID_Found(SkyrimMod mod)
```
Noggog: There's no way you would know this, as it's tooling I wrote to behave how I want, but I've set up the Mutagen autofixture systems to auto add records when written like this:
`TryResolveKeyword_ByEditorID_Found(SkyrimMod mod, Npc npc, Keyword keyword)`

Both npc and keyword will be added to the mod,   Stuff like `npc.Keyworsd.Add(keyword)` is still needed, though.   Just the adding to the SkyrimMod is "automated"

Mutagen.Bethesda.Core/Plugins/Aspects/IKeyworded.cs
MarkKoz: When using `var` it chooses a suboptimal overload apparently:
```
CS0618: Method 'Mutagen.Bethesda.Plugins.Cache.ILinkCache.TryResolve(FormKey, out IMajorRecordGetter, ResolveTarget)' is obsolete: This call is not as optimized as its generic typed counterpart. Use as a last resort.
```

Mutagen.Bethesda.UnitTests/Plugins/Aspects/IKeywordedTests.cs
MarkKoz: Thanks - will keep that in mind!

Mutagen.Bethesda.Core/Plugins/Aspects/IKeyworded.cs
Noggog: Oh, then maybe this:
`if (cache.TryResolve<TKeyword>(keywordForm, out var keyword))`

Noggog (approved):

Mutagen.Bethesda.Core/Plugins/Aspects/IKeyworded.cs
MarkKoz: Ah yeah, that did it. So `TKeyword?` is fine, but it has to be used with `TryResolve<TKeyword>`. Otherwise, it uses ` TryResolve<TMajor>`, and that does not like `TKeyword?`.

Noggog (approved):

PR #472, opened 2023-08-26
Define GlobalGroupMixIn and GameSettingGroupMixIn for Fallout 4

### LINKED ISSUE #467: Define GameSettingGroupMixIn for Fo4
Mostly copy paste the existing setups in Skyrim/Oblivion

MarkKoz:
Closes #467 

Implementing `GlobalGroupMixIn` required an overhaul to the Global record. This brings its API into alignment with Oblivion's and Skyrim's Global record. At one point (https://github.com/Mutagen-Modding/Mutagen/commit/21888983b9f7f2afbbc318c1d8e5250106fa350d) we had this for Fallout 4 but then reverted it for unclear reasons. After some discussion in Discord, we concluded that it seems fine to add this back in; the passthrough tests should tell us whether this change is safe.

Speaking of which, I have not run the passthrough tests yet, so I'm marking this as a draft until we can confirm these changes pass tests.


Noggog:
I think the change that was confusing and inspired the alternative structure, was that FNAM the type character subrecord was now optional, where it's assumed float if it's missing?

So the GlobalUnknown was no longer needed, but an extra tracker of whether to output the FNAM was needed on the float, since it may or may not export that

PR #476, opened 2023-09-02
Fixes list rendering in documentation

ThePoShWolf:
I'm very excited to have found this library! After spending some time reading the documentation, I have found that lists in the Mutagen documentations aren't rendered properly in GitHub Pages, for example in [index.md](https://mutagen-modding.github.io/Mutagen/):

image

This PR has a very simple fix to add a preceding empty line before all markdown lists in the docs/ folder to ensure they are rendered as lists.

Accomplished with a simple regex replace:

```plaintext
'^([^-].*)\r\n-', '$1

-'
```

### DIFF
--- a/docs/best-practices/Accessing-Known-Records.md
+++ b/docs/best-practices/Accessing-Known-Records.md
@@ -33,6 +33,7 @@ First, a FormKey is created pointing to the known record for the Argonian race i
 This will retrieve the winning override.
 
 However, there's a few annoyances:
+
 - Neither the dev nor future readers know the FormID for records offhand, and so must always look them up.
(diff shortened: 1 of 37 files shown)


Noggog (approved):
Thanks a ton!  For whatever reason, Github's wiki systems didn't mind the old formatting, and I didnt notice the issue when migrating to the newer doc setup.

Thanks for the fix!

Noggog:
Hmm, unfortunately, didn't seem to fix it on the actual site.  Maybe there's something specific about mkDocs that are causing issues

EDIT: Nevermind.  The homepage index was just missed, so it was still displaying messed up lists.   Fixed that one as a followup

Issue #469, opened 2023-08-21
Upgrade Mutagen.Bethesda.Testing.GUI to bootstrap itself if no settings file exists


Noggog:
Right now there's no buttons to add the initial left hand side game sections
![image](https://github.com/Mutagen-Modding/Mutagen/assets/24981326/7b8d8c8d-69ee-441d-af11-5e9198568735)

Noggog:
Here's my test settings file, to get people up and running until this is done:
[TestingSettings.zip](https://github.com/Mutagen-Modding/Mutagen/files/12393616/TestingSettings.zip)

Unzip and place the json file inside `[Repo]/Mutagen.Bethesda.Tests` folder (if running the test UI from the IDE)
![image](https://github.com/Mutagen-Modding/Mutagen/assets/24981326/bea85be1-cc7e-46c8-9338-036b178e42e1)

PR #483, opened 2023-09-18
Implement #469

Elscrux:
@Noggog Let me know if this works like you intended. Basically, it creates the settings with all the information it has and it should properly handle existing files too.


Noggog (approved):
fixes #469

PR #484, opened 2023-09-18
feat: starfield PTST, AOPF, DEBR, AVMD, OSWP parsed

FlayaN:
- PTST (only added to starfieldmod.xml due to only containing a EDID)
- AOPF renamed to AudioOcclusionPrimitive and added float instead of bytearray to last two props (still missing good name, not sure how to find this, so still named as the recordtype for now)
- DEBR parsed
- AVMD parsed
  - Also added SkinFormComponent and BodyPartInfoComponent in Component.xml with required changes needed to support it.
- OSWP (only added to starfieldmod.xml due to only containing a EDID)


Mutagen.Bethesda.Starfield/Records/Major Records/ActorValueModulation.cs
```diff
@@ -0,0 +1,12 @@
+namespace Mutagen.Bethesda.Starfield;
+
+public partial class ActorValueModulation
+{
+    public enum TypeEnum
+    {
+        UNKNOWN, // todo '' in xedit
```
Noggog: In Mutagen, we omit enum entries we don't know, as they can always be accessed still by interacting with the enum as an `int`

Mutagen.Bethesda.Starfield/Records/Major Records/Debris.xml
```diff
@@ -2,6 +2,20 @@
...
+      <RefList name="Models" refName="DebrisModel" />
+    </Fields>
+  </Object>
+  <Object name="DebrisModel" objType="Subrecord" GUID="1135d2a6-9241-42d7-a689-3ac8efe03b0b" ID="902">
+    <Fields>
+      <Data>
+        <Fields>
+          <UInt8 name="Percentage" />
+          <AssetLink name="ModelFilename" binaryType="NullTerminate" assetType="StarfieldModel" />
+          <Break />
```
Noggog: `<Break />` is added when the fields following it are only present on "newer" versions of the record.   The more official way to handle it is to identify the FormVersion that the fields in question get added, and then add the versioning instructions for those.   But often, that's annoying/impossible, and so a `<Break/>` is just added to say "dunno what version it is, but sometimes things after this are missing".

However, when new games roll around, they often dont have the "older" version with the missing fields, so these breaks can be removed.  Always good to test without them to see if they're still needed.   I think in this case it isn't?

Mutagen.Bethesda.Starfield/Records/Major Records/ActorValueModulation.xml
```diff
@@ -0,0 +1,21 @@
+<?xml version="1.0" encoding="us-ascii"?>
+<Loqui xmlns="http://tempuri.org/LoquiSource.xsd">
+  <Object name="ActorValueModulation" recordType="AVMD" objType="Record" baseClass="StarfieldMajorRecord" GUID="15904907-62a1-4cb0-859a-00c52c7cacdc" ID="740">
+    <Fields>
+      <RefList name="Components" refName="AComponent"/>
+      <Enum name="Type" enumName="ActorValueModulation.TypeEnum" byteLength="4" recordType="MNAM" nullable="False" />
+      <String name="YNAM" recordType="YNAM" />
+      <String name="Name" recordType="TNAM" nullable="False" />
+      <UInt32 name="ITMC" recordType="ITMC" nullable="False" />
```
Noggog: I think the xEdit defs are being updated quickly, so this might not have been labeled at the time you looked.  But even so, often before a list is a subrecord that is the counter, which this is.

So we can remove the `ITMC` field here, and just put `counterRecType="ITMC"` on the list field

Mutagen.Bethesda.Starfield/Records/Major Records/ActorValueModulation.xml
```diff
@@ -0,0 +1,21 @@
...
+      <RefList name="ActorValueModulationSubrecords" refName="ActorValueModulationSubrecord" />
+      <UInt32 name="MODT" recordType="MODT" />
+    </Fields>
+  </Object>
+  <Object name="ActorValueModulationSubrecord" objType="Subrecord" GUID="175f1795-017c-4036-b881-cf923d19d435" ID="903">
+    <Fields>
+      <String name="LNAM" recordType="LNAM" nullable="False" />
+      <String name="VNAM" recordType="VNAM" />
+      <ByteArray name="Color" recordType="NNAM" unknown="true" />
```
Noggog: I think this can be swapped to a <Color> record, with the attribute `binaryType="Alpha"`  Give it a go and see

Mutagen.Bethesda.Starfield/Records/Major Records/ActorValueModulation.xml
```diff
@@ -0,0 +1,21 @@
...
+<Loqui xmlns="http://tempuri.org/LoquiSource.xsd">
+  <Object name="ActorValueModulation" recordType="AVMD" objType="Record" baseClass="StarfieldMajorRecord" GUID="15904907-62a1-4cb0-859a-00c52c7cacdc" ID="740">
+    <Fields>
+      <RefList name="Components" refName="AComponent"/>
+      <Enum name="Type" enumName="ActorValueModulation.TypeEnum" byteLength="4" recordType="MNAM" nullable="False" />
+      <String name="YNAM" recordType="YNAM" />
+      <String name="Name" recordType="TNAM" nullable="False" />
+      <UInt32 name="ITMC" recordType="ITMC" nullable="False" />
+      <RefList name="ActorValueModulationSubrecords" refName="ActorValueModulationSubrecord" />
+      <UInt32 name="MODT" recordType="MODT" />
```
Noggog: I think this enum has been defined in xEdit

BTW, dunno if it's needed for this one, but there's tooling in https://github.com/Mutagen-Modding/Mutagen.Bethesda.Generation.Tools to transform xEdit enums to C# ones (for when there's hundreds of entries, haha)

Noggog (changes requested):

Mutagen.Bethesda.Starfield/Records/Major Records/ActorValueModulation.cs
FlayaN: Ah nice! Removed that entry

Mutagen.Bethesda.Starfield/Records/Major Records/Debris.xml
FlayaN: Ah, good info! Thanks, removed it. Still passing without it, so not needed

Mutagen.Bethesda.Starfield/Records/Major Records/ActorValueModulation.xml
FlayaN: Really good info, worked perfectly with it merged into the list, also renamed the subrecord to match latest xedit

Mutagen.Bethesda.Starfield/Records/Major Records/ActorValueModulation.xml
FlayaN: Ah nice, yes you were correct! It was a `<Color />` with the `binaryType="Alpha"`

Mutagen.Bethesda.Starfield/Records/Major Records/ActorValueModulation.xml
FlayaN: Yep, wasn't there when I added this :D Updated to use it now, also setup the tools to run it. Had to check the code what part of the xedit it expected, but figured out that the input file had to only contain the

```
          {0} 'None',
          {1} 'Rough',
          {2} 'Opacity',
          {3} 'Normal',
          {4} 'AO',
          {5} 'Metal',
          {6} 'Conditions',
          {7} 'Emissive'
```

And then generating

```
public enum EnumName
{
    None = 0,
    Rough = 1,
    Opacity = 2,
    Normal = 3,
    AO = 4,
    Metal = 5,
    Conditions = 6,
    Emissive = 7,
}
```

Like you said in this case it was probably not needed to run it through, but nice to know how and when to use it 👍

Mutagen.Bethesda.Starfield/Records/Major Records/ActorValueModulation.xml
```diff
@@ -0,0 +1,20 @@
+<?xml version="1.0" encoding="us-ascii"?>
+<Loqui xmlns="http://tempuri.org/LoquiSource.xsd">
+  <Object name="ActorValueModulation" recordType="AVMD" objType="Record" baseClass="StarfieldMajorRecord" GUID="15904907-62a1-4cb0-859a-00c52c7cacdc" ID="740">
+    <Fields>
+      <RefList name="Components" refName="AComponent"/>
+      <Enum name="Type" enumName="ActorValueModulation.TypeEnum" byteLength="4" recordType="MNAM" nullable="False" />
+      <String name="YNAM" recordType="YNAM" />
+      <String name="Name" recordType="TNAM" nullable="False" />
+      <RefList name="ActorValueModulationEntries" refName="ActorValueModulationEntry" counterRecType="ITMC" />
```
Noggog: One last "convention" thing I just noticed:
While it's sometimes required to name classes somewhat verbosely, like `ActorValueModulationEntry` (which is fine), member names can often be shorter.  Like in this case, something like `Entries` probably works just fine
`myAvMod.Entries.ForEach(...)`

FlayaN: Ah ofc! Renamed to Entries

PR #485, opened 2023-09-22
Starfield: Implement OTFT, LTEX, ANIO, SCOL, AMDL, AORU, BNDS, DFOB and LGDI

Elscrux:


Mutagen.Bethesda.Starfield/Records/StarfieldMod.xml
```diff
@@ -23,6 +23,14 @@
       <Group name="Debris" refName="Debris" />
       <Group name="ActorValueModulations" refName="ActorValueModulation" />
       <Group name="BiomeSwaps" refName="BiomeSwap" />
+      <Group name="Outfit" refName="Outfit" />
```
FlayaN: These names should be plural

Elscrux: Good point, thanks!

Mutagen.Bethesda.Starfield/Records/Major Records/LandscapeTexture.xml
```diff
@@ -2,6 +2,15 @@
...
     <Fields>
+      <AssetLink name="MaterialPath" recordType="BNAM" assetType="StarfieldMaterial" nullable="True" />
+      <FormLink name="MaterialType" recordType="MNAM" refName="MaterialType" nullable="False" />
+      <Data recordType="HNAM">
+        <Fields>
+          <UInt8 name="HavokFriction" />
+          <UInt8 name="HavokRestitution" />
+        </Fields>
+      </Data>
+      <Float name="UnknownQNAM" recordType="QNAM" unknown="True" />
```
Noggog: Not worth caring too much, but been naming unknown fields like this just `QNAM`

Mutagen.Bethesda.Starfield/Records/Major Records/LegendaryItem.xml
```diff
@@ -2,6 +2,28 @@
...
+      <FormLink name="ApplicableItemList" refName="LeveledItem" recordType="ANAM" />
+      <FormLink name="LegendaryTemplateList" refName="LegendaryItem" recordType="ENAM" />
+      <RefList name="LegendaryMods" refName="LegendaryMod" recordType="BNAM" />
+      <RefList name="IncludeFilters" refName="LegendaryFilter" recordType="CNAM" />
+      <RefList name="ExcludeFilters" refName="LegendaryFilter" recordType="DNAM" />
+    </Fields>
+  </Object>
+  <Object name="LegendaryMod" objType="Subrecord">
+    <Fields>
+      <Enum name="Flags" enumName="LegendaryItem.StarSlot" />
```
Noggog: Let's name this field `Slot`.   `Flags` makes more sense if it had several misc flags, which this isn't.

Almost feels like this could just be a number, except for the fact that it would be zero indexed and off by one, then.

Mutagen.Bethesda.Starfield/Records/Major Records/StaticCollection.xml
```diff
@@ -1,7 +1,35 @@
...
+    </Fields>
+  </Object>
+  <Object name="StaticPart" objType="Subrecord">
+    <Fields>
+      <Data recordType="ONAM">
+        <Fields>
+          <FormLink name="Static">
+            <Interface>IStaticTarget</Interface>
+          </FormLink>
+          <ByteArray name="Unknown" byteLength="4" nullable="false" unknown="true" />
```
Noggog: Let's swap this to Int32.   If we know the length with some certainty, it's better to not use arrays

Noggog (changes requested):

Mutagen.Bethesda.Starfield/Records/Major Records/LegendaryItem.xml
Elscrux: Totally agree on the name! I think it might be a value that is limited to a certain range so enum might be better?

Noggog: Ints have the ability to have min/max values, but the zero index issue is reason enough to keep it as enum, I think?  Definitely an odd field

Noggog (approved):

PR #486, opened 2023-09-25
Add Starfield weather definition (based on Fallout 4, not a lot changed)

Elscrux:


Mutagen.Bethesda.Starfield/Records/Major Records/Weather.xml
```diff
@@ -2,6 +2,173 @@
...
+          <RefDirect name="SkyUpperColor" refName="WeatherColor" />
+          <RefDirect name="FogNearColor" refName="WeatherColor" />
+          <RefDirect name="UnknownColor" refName="WeatherColor" />
+          <RefDirect name="AmbientColor" refName="WeatherColor" />
+          <RefDirect name="SunlightColor" refName="WeatherColor" />
+          <RefDirect name="SunColor" refName="WeatherColor" />
+          <RefDirect name="StarsColor" refName="WeatherColor" />
+          <RefDirect name="SkyLowerColor" refName="WeatherColor" />
+          <RefDirect name="HorizonColor" refName="WeatherColor" />
+          <Break />
```
Noggog: Let's see if we can remove the Breaks since we're on a new game

Mutagen.Bethesda.Starfield/Records/Major Records/WeatherAmbientColorSet.cs
```diff
@@ -0,0 +1,107 @@
+namespace Mutagen.Bethesda.Starfield;
+
+partial class WeatherAmbientColorSet
+{
+    // IAmbientColorsGetter IWeatherAmbientColorSetGetter.this[TimeOfDay time] => time switch
```
Noggog: Can these be uncommented out?

Noggog (changes requested):

Mutagen.Bethesda.Starfield/Records/Major Records/WeatherAmbientColorSet.cs
Elscrux: This is done because Starfield doesn't have this `TimeOfDay` enum. Instead, there is some kind of `TimeOfDayData` that hasn't been decoded yet, so there is not much to be done I think.

Mutagen.Bethesda.Starfield/Records/Major Records/Weather.xml
Elscrux: Ok, I'll try. What do those Breaks to btw?

Noggog: Some fields are added as the FormVersion increases.  Rather than figuring out the exact formversion that it's added (sometimes impossible to know), the Break is a way to just say "I know content past here might be missing occasionally, so be aware"

With a newer game though, I imagine all the fields from the past game should always be present, so the breaks from old definitions probably arent needed

Elscrux: I had to fix some bugs that I somehow didn't see before, but now it also works without breaks :)

PR #489, opened 2023-10-03
Add alpha layer data definition for landscape

Elscrux:


Elscrux:
I couldn't test it properly via the testing framework because of some problems running the tests. But the data this produces looks fine.

Noggog (approved):

PR #495, opened 2023-12-08
Fix mismatching strings entry length

Elscrux:
Fix mismatching length for encodings where the byte count and string length are different.

I tested this for the Dragonborn book DLC2Book2CommonRavenRock01. Using non-latin languages like Japanese or Russian had their book text cut off roughly in the middle because it was using the string length, not the byte length of the encoding that is used.


Noggog:
Thanks!   Would be nice to have unit tests to back the change, but no worries if you don't have the time.   Happy holidays!

Issue #510, opened 2024-04-01
MajorRecord type enumerator


Noggog:
Add a way to easily enumerate all major record types in a given game release

Right now the easiest way is to iterate `LoquiRegistration.StaticRegister.Registrations` and filter on any that implement IMajorRecordGetter.   Would be good to add a more mutagen official route, though

Elscrux:
Where would you like to have this? In the Core project in some folder?

Noggog:
Maybe within `Core/Plugins/Records/Mapping`?  That's where logic for finding what link interfaces exist given a GameCategory, or looking up the getter type from any mutagen class type, etc.   Seems like a decent spot to pop in a `MajorRecordTypeEnumerator` that has a `IEnumerable<Type> GetMajorRecordTypesFor(GameCategory cat)`

Noggog:
For now it can just house the LoquiRegistration loop in a lazy variable?  Maybe eventually we'll hyperoptimize and have the code generation fish for all the types ahead of time

PR #513, opened 2024-04-02
Add MajorRecordTypeEnumerator

Elscrux:
Implements #510 

@Noggog Let me know if this is what you were thinking or if I should change anything.

Btw, could you expand the .editorconfig with the guidelines you use for Mutagen. I keep putting the curly braces in the wrong place lol.

### DIFF
--- a/Mutagen.Bethesda.Core/Plugins/Records/Mapping/MajorRecordTypeEnumerator.cs
+++ b/Mutagen.Bethesda.Core/Plugins/Records/Mapping/MajorRecordTypeEnumerator.cs
@@ -1,4 +1,5 @@
-﻿using Loqui;
+﻿using System.Collections.Concurrent;
+using Loqui;
 using Noggog;
 namespace Mutagen.Bethesda.Plugins.Records.Mapping;
 
@@ -6,7 +7,7 @@ public record RecordTypes(Type ClassType, Type GetterType, Type SetterType);
 
 public static class MajorRecordTypeEnumerator
 {
-	private static readonly Dictionary<GameCategory, List<ILoquiRegistration>> Registrations = new();
+	private static readonly ConcurrentDictionary<GameCategory, List<ILoquiRegistration>> Registrations = new();
 
 	private static List<ILoquiRegistration> GetRegistrations(GameCategory cat)
 	{
--- a/Mutagen.Bethesda.Core/Plugins/Records/Mapping/MajorRecordTypeEnumerator.cs
+++ b/Mutagen.Bethesda.Core/Plugins/Records/Mapping/MajorRecordTypeEnumerator.cs
@@ -6,7 +6,7 @@ public static class MajorRecordTypeEnumerator
 {
 	private static readonly Dictionary<GameCategory, List<ILoquiRegistration>> Registrations = new();
 
-	public static List<ILoquiRegistration> GetRegistrations(GameCategory cat)
+	private static List<ILoquiRegistration> GetRegistrations(GameCategory cat)
 	{
 		var gameNamespace = "Mutagen.Bethesda." + cat;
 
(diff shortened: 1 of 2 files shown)


Mutagen.Bethesda.Core/Plugins/Records/Mapping/MajorRecordTypeEnumerator.cs
```diff
@@ -0,0 +1,42 @@
+﻿using Loqui;
+using Noggog;
+namespace Mutagen.Bethesda.Plugins.Records.Mapping;
+
+public static class MajorRecordTypeEnumerator
+{
+	private static readonly Dictionary<GameCategory, List<ILoquiRegistration>> Registrations = new();
```
Noggog: Right now this is vulnerable to multithreading collisions, as many threads can call the accessor functions and all be filling this dictionary/list in parallel

I'd modify this to a `Lazy<IReadOnlyDictionary<....>>` with the GetRegistrations logic as the callback to fulfill the lazy variable

The public accessors can then just access the lazy variable and do the lookup.  The lazy variables have thread safety mechanics internally

Mutagen.Bethesda.Core/Plugins/Records/Mapping/MajorRecordTypeEnumerator.cs
```diff
@@ -0,0 +1,42 @@
...
+
+	private static List<ILoquiRegistration> GetRegistrations(GameCategory cat)
+	{
+		var gameNamespace = "Mutagen.Bethesda." + cat;
+
+		return LoquiRegistration.StaticRegister.Registrations
+			.Where(x =>
+				string.Equals(x.ClassType.Namespace, gameNamespace, StringComparison.Ordinal)
```
Noggog: Might break if we change namespaces (not that we would).  Maybe downgrade it to Contains(..) on the game category string?

Let's make sure to turn the game category into a string outside the loop (like this PR is currently doing), or use `Enums<GameCategory>.ToStringFast(...)`, as ToStringing enums is actually heavyier than expected

Might be a better way besides string checking namespaces, but not coming to my brain offhand

Mutagen.Bethesda.Core/Plugins/Records/Mapping/MajorRecordTypeEnumerator.cs
```diff
@@ -0,0 +1,42 @@
...
+	{
+		return Registrations.GetOrAdd(cat, () => GetRegistrations(cat))
+			.Select(x => x.GetterType);
+	}
+
+	public static IEnumerable<Type> GetMajorRecordSetterTypesFor(GameCategory cat)
+	{
+		return Registrations.GetOrAdd(cat, () => GetRegistrations(cat))
+			.Select(x => x.SetterType);
+	}
```
Noggog: I think we can unify this into one call, and expose a record like `InterfaceMappingTypes`, but maybe call it `RecordTypes` and have Class/Getter/Setter as required fields?

Then we can just have one function that returns this record, and the user can get whatever subtype they desire from there

Noggog (changes requested):

Mutagen.Bethesda.Core/Plugins/Records/Mapping/MajorRecordTypeEnumerator.cs
Elscrux: Wouldn't a ConcurrentDictionary also work?

Elscrux: The lazy variant would have to fill the whole directory for all categories when it's first called right? Otherwise it would have the same problem as the dictionary has now.

Noggog: Yes, if you used a lazy variable straight up it would have to do all the game categories up front to make sense.
you could do a concurrent dictionary as well if you wanted to go that route

Elscrux: I used the ConcurrentDictionary as it seemed easiest and most efficient :)

Noggog (approved):

Issue #474, opened 2023-08-28
AssetLink Equality


Noggog:
Model's file member is not checking equality properly in Mutagen's generated equality functionality

```
 lhs.File
  {Meshes\\DLC02\\Armor\\StahlrimHeavy\\HelmetGND.nif}
    DataRelativePath: "Meshes\\DLC02\\Armor\\StahlrimHeavy\\HelmetGND.nif"
    Extension: ".nif"
    IsNull: false
    RawPath (Mutagen.Bethesda.Plugins.Assets.AssetLinkGetter<Mutagen.Bethesda.Skyrim.Assets.SkyrimModelAssetType>): "DLC02\\Armor\\StahlrimHeavy\\HelmetGND.nif"
    RawPath: "DLC02\\Armor\\StahlrimHeavy\\HelmetGND.nif"
    Type: {Mutagen.Bethesda.Skyrim.Assets.SkyrimModelAssetType}
    _rawPath: "DLC02\\Armor\\StahlrimHeavy\\HelmetGND.nif"

  rhs.File
  {Meshes\\DLC02\\Armor\\StahlrimHeavy\\HelmetGND.nif}
    DataRelativePath: "Meshes\\DLC02\\Armor\\StahlrimHeavy\\HelmetGND.nif"
    Extension: ".nif"
    IsNull: false
    RawPath: "DLC02\\Armor\\StahlrimHeavy\\HelmetGND.nif"
    Type: {Mutagen.Bethesda.Skyrim.Assets.SkyrimModelAssetType}
    _rawPath: "DLC02\\Armor\\StahlrimHeavy\\HelmetGND.nif"
```

is returning false

PR #515, opened 2024-04-04
Asset Link Equality

Elscrux:
Fixes issues reported in #474 and more.
The mentality on the equality operations was to report functionally equals paths as equal, which is why I opted to use DataRelativePath for all equality operations (this was inconsistent between DataRelativePath and RawPath before).


Noggog:
Closes #474

Noggog (approved):

PR #545, opened 2024-08-24
Migrate from DotNetZip to `System.IO.Compression`.

focustense:
DotNetZip is deprecated and its chain of transitive dependencies leads to a version of `System.Drawing.Common` with a known security vulnerability.

`System.IO.Compression` has mostly the same APIs except that `Read` methods are not exact; instead, the newer `Stream.ReadExact` method needs to be used to guarantee full decompression.


Noggog:
Looking good!  Thanks!
image

Noggog (approved):

PR #547, opened 2024-08-29
Add WithAllParentMasters binary write builder option

Elscrux:


Mutagen.Bethesda.Core/Plugins/Binary/Translations/BinaryWriteBuilder.cs
```diff
@@ -1375,18 +1378,42 @@ public FileBinaryModdedWriteBuilder<TModGetter> WithExplicitOverridingMasterList
...
+
+        return this with
+        {
+            _params = _params with
+            {
+                _param = _params._param with
+                {
+                    ExtraIncludeMasters = _params._param.ExtraIncludeMasters.EmptyIfNull().And(sortedMasters).Distinct().ToArray()
+                }
+            }
```
Noggog, line 1412: Can we refactor this logic to an internal component that takes in the `LO` and returns the `IEnumerable<ModKey>`?

Once it's its own component, it'll be easier to test correctness, too, etc.

Main thing is that this builder call shouldn't be the one specifying the LO as an input parameter.  The LO is set via other builder calls.  Since not all the info is known at the time of THIS call, the logic written here should be run "lazily" at the end once all the bits are set and the Write is actually called

To get the "laziness", a system like the `_loadOrderSetter` will need to be added like a `_mastersContentSetter`.   When called during the "pull together" phase, it would pull in the LO set right before, and then use that to calculate the masters.   I can do this part if you want, as it's pretty fiddly and error prone

Noggog (changes requested):

Mutagen.Bethesda.Core/Plugins/Binary/Translations/BinaryWriteBuilder.cs
Elscrux: I looked at the `_loadOrderSetter` before, but the load order it has just has a limited `IModFlags` interface or so, which didn't allow master references to be looked at. I didn't want to edit the existing builder code so I went this way for now.

Noggog: True true.  hmm

Well.   I think the idea is that load order is mainly used for ordering.  Whereas the Data folder is used for looking up info about mods.  So the presence of `ILoadOrderGetter<IModFlagsGetter> loadOrder` is maybe deceiving?

I think someone like yourself using `WithAllParentMasters()` would:
- not supply anything in the `WithAllParentMasters()` call itself
- Would include `WithDataFolder(...)` to let the system know where to read mods from to find all the parent masters
- Would include `WithLoadOrder(...)` to let the system know how to order all the parent masters

Maybe as a followup, we could remove the `ILoadOrderGetter<IModFlagsGetter> loadOrder`, but I would want to deep dive on it to make sure that's a good cut.

PR #561, opened 2024-10-16
changed mkdoc tab color to increase visibility

some1one:
changed mkdoc tab color to increase visibility, feel free to change the colors

### DIFF
--- a/docs/Contributing.md
+++ b/docs/Contributing.md
@@ -45,11 +45,12 @@ This step downloads (clones) the fork you just made to your local computer.
 ![Cloning](images/cloning.png){ width="350" }
 
 #### Installing MkDocs
-This is the documentation system Mutagen uses, so you have to install it so it can display the docs as you edit them.
+This is the documentation system Mutagen uses, so you have to install it, along with the GLightbox plugin, so it can display the docs as you edit them.
 
 [Installing Material for MkDocs](https://squidfunk.github.io/mkdocs-material/getting-started/)
+[Installing MkDocs GLightbox](https://blueswen.github.io/mkdocs-glightbox/)
 
-TLDR: `pip install mkdocs-material` in a command line
+TLDR: `pip install mkdocs-material mkdocs-glightbox` in a command line
 
 #### Running MkDocs
 Running `mkdocs serve` on your cloned Mutagen repository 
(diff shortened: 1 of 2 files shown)


Noggog (changes requested):
Seems like it's behaving oddly where they're all orange if the first tab is selected, and all not orange if any other tab is selected

NVIDIA_Share_eYALcy8eta

some1one:
was a copy paste typo, fixed

some1one:
however, the background color of the unselected tabs is supposed to be a different color as well. that part is working for me so im not sure why its not showing up on yours
`.tabbed-labels > label {
  background-color: (219.29deg 73.01% 27.85% / 64%)
}`

Noggog:
> the background color of the unselected tabs is supposed to be a different color as well. that part is working for me so im not sure why its not showing up on yours

Yeah, not sure why either.   Maybe the official deployment will do it right?

Also, I'm not a CSS guy, so maybe this is obvious, but how do i interpret `background-color: (219.29deg 73.01% 27.85% / 64%)` to be orange?  

Asking as one thing I'd like to try while we're here is potentially making the active tab's text color to be the same bright "salmon orange" here:
image
(#fa7376)

If we did that, the background color might need tweaking to a darker color, or maybe purple to give the Mutagen color effect
(#372773)
But not sure how to start experimenting with that offhand since i dont understand the background color, haha

----------

Approved for now.  Some of the experiments above are just musing/followup ideas.  Will merge this in as-is if those fizzle soon

Noggog (approved):

some1one:
The background color is just the format the original background color was in and i just used the browser color picker to alter it so i guess it kept the same format. it probably will show the hexadecimal version somewhere we could switch to.

I do like the text color idea and changing the active background color to another shade. instead of the whole tab being orange.

some1one:
okay, so this is the updated colors along with the blue background mine shows for the inactive tabs. aside from the blue background, the colors use the theme variables instead of the hex codes.

image

some1one:
actually, I found a theme color that's pretty close for the blue too. might be too light though.
image

some1one:
i also added a drop shadow. it looks pretty nice now and will update with the theme colors.

some1one:
rebased

Noggog:
Thanks for all the extra experimenting!

PR #563, opened 2024-10-17
Top Level Groups Enumeration

some1one:
+added WhereNotNull IEnumerable Extension
+added GetTopLevelMajorRecordTypes to MajorRecordTypeEnumerator
+added to cheatsheat
*changed cache list type to immutableArray

### DIFF
--- a/Mutagen.Bethesda.Core/Extensions/EnumerableExt.cs
+++ b/Mutagen.Bethesda.Core/Extensions/EnumerableExt.cs
@@ -4,6 +4,11 @@ namespace Mutagen.Bethesda;
 
 internal static class EnumerableExt
 {
+    public static IEnumerable<T> WhereNotNull<T>(this IEnumerable<T?> enumer)
+    {
+        return enumer.Where(x => x != null).Cast<T>();
+    }
+
     public static ExtendedList<T> CastExtendedList<T>(this ExtendedList<T> enumer)
     {
         return enumer;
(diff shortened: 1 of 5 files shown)


some1one:
rebased on dev and removed the doc commits from the other pull request

Mutagen.Bethesda.Core/Extensions/EnumerableExt.cs
```diff
@@ -4,6 +4,11 @@ namespace Mutagen.Bethesda;
 
 internal static class EnumerableExt
 {
+    public static IEnumerable<T> WhereNotNull<T>(this IEnumerable<T?> enumer)
```
Noggog, line 7: Already got an extension like this over in https://github.com/Noggog/CSharpExt/blob/89a5e0918ae8a212b592b890a5b62ca1209f2120/Noggog.CSharpExt/Extensions/EnumerableExt.cs#L257

But I do like the `Where__` naming better.   Feel free to open a PR over there, and mark the old ones obsolete

Noggog (changes requested):
Rebase went a bit weird and some commits got in both, but all good!
image

Only thing I'd say is maybe (re)move the null extension, and should be good to go!

Noggog:
Ah, i can handle the small tweaks.   Like i mentioned, gonna remove the new WhereNotNull method, but feel free to add to CSharpExt if you feel inspired to do so

Noggog (approved):

PR #567, opened 2024-11-16
Condition Parameter Generation

Elscrux:
Generate Parameter1 and Parameter2 properties to enable access to parameters without knowing the concrete condition data type


Mutagen.Bethesda.Skyrim/Records/Common Subrecords/CanFlyHereConditionData_Generated.cs
```diff
@@ -454,6 +454,25 @@ public static implicit operator TranslationMask(bool defaultOn)
         }
         #endregion
 
+        #region Mutagen
+        public object? Parameter1
+        {
+            get => null;
```
Noggog, line 460: Getting null here feels fine.  Feels a bit odd to just do nothing in the set.   Gut is telling me we should throw if the given value is not null (similar to how we throw if the type is wrong for used parameters)

Mutagen.Bethesda.Skyrim/Records/Common Subrecords/GetCrimeGoldViolentConditionData_Generated.cs
```diff
@@ -466,6 +466,19 @@ public static implicit operator TranslationMask(bool defaultOn)
         #region Mutagen
         public override IEnumerable<IFormLinkGetter> EnumerateFormLinks() => GetCrimeGoldViolentConditionDataCommon.Instance.EnumerateFormLinks(this);
         public override void RemapLinks(IReadOnlyDictionary<FormKey, FormKey> mapping) => GetCrimeGoldViolentConditionDataSetterCommon.Instance.RemapLinks(this, mapping);
+        public object? Parameter1
+        {
+            get => Faction;
+            set => Faction = (value is IFormLinkOrIndex<IFactionGetter> v ? v : throw new ArgumentException());
```
Noggog, line 472: I like the throw if the type is wrong.

Might be nice to add `Type Parameter1Type { get; }` parameter so that the user can know/expect what type should be given?

Mutagen.Bethesda.Generation/Modules/Plugin/ConditionDataModule.cs
```diff
@@ -0,0 +1,58 @@
...
+            sb.AppendLine($"public object? Parameter{i}");
+            using (sb.CurlyBrace())
+            {
+                if (field.Name.Contains("Unused"))
+                {
+                    sb.AppendLine("get => null;");
+                    sb.AppendLine("set");
+                    using (sb.CurlyBrace())
+                    {
+                        sb.AppendLine();
```
Noggog, line 44: Putting it here too, but we shuld check if it's not null and throw argument exception if it's not null?

Mutagen.Bethesda.Generation/Modules/Plugin/ConditionDataModule.cs
```diff
@@ -0,0 +1,58 @@
...
+        if (obj.BaseClassName != "ConditionData") return;
+        if (obj.GetObjectData().GenerateConditionData == false) return;
+
+        var i = 1;
+        foreach (var field in obj.Fields)
+        {
+            if (i > 2) break;
+            if (field.Name.Contains("UnusedStringParameter")) continue;
+
+            sb.AppendLine($"public object? Parameter{i}");
```
Noggog: Let's add `public Type? Parameter1Type` here as well

Mutagen.Bethesda.Skyrim/Records/Common Subrecords/ConditionData.cs
```diff
@@ -19,11 +19,29 @@ partial class ConditionData : IFormLinkOrIndexFlagGetter, IConditionStringParame
     string? IConditionStringParameterGetter.FirstStringParameter => throw new NotImplementedException();
     string? IConditionStringParameterGetter.SecondStringParameter => throw new NotImplementedException();
     Condition.Function IConditionDataGetter.Function => throw new NotImplementedException();
+    public object? Parameter1
```
Noggog: Defining these like this makes it harder for the implementing classes to implement the actual setups.

In theory, these could be `virtual`, which would mean the inheriting classes would need to `override`.
that's one way.

The way I've typically been doing it is writing them like the lines above, which specifically associate them with the interface call, rather than being straight fields on the class

I think this allows the inheriting classes to define the fields for real themselves for their implementation of the interface.

Noggog (changes requested):
On a good track.  Few more things to add, i think!

Mutagen.Bethesda.Skyrim/Records/Common Subrecords/CanFlyHereConditionData_Generated.cs
Elscrux: I was worrying this could lead to strange behavior when you are working on generic conditions. Setting a property might not be intuitively something where you expect it throws when doing it? What do you think? You can of course just add a try catch block when you get any issues with this.

Mutagen.Bethesda.Skyrim/Records/Common Subrecords/GetCrimeGoldViolentConditionData_Generated.cs
Elscrux: Yeah that's a good idea

Mutagen.Bethesda.Generation/Modules/Plugin/ConditionDataModule.cs
Elscrux: I think this is a similar design decision as throwing the other exception. Should be both or none.

Mutagen.Bethesda.Skyrim/Records/Common Subrecords/ConditionData.cs
Elscrux: I implemented the interfaces explicitly. I think it might be good if we didn't implement the properties explicitly in ConditionData so we enforce that concrete types implement it (we could make them abstract for example).

Mutagen.Bethesda.Skyrim/Records/Common Subrecords/CanFlyHereConditionData_Generated.cs
Noggog: Yeah, not a great way either route... but I think that's just the price of writing generic API that is trying to handle everything.

The downside of the current setup is that users will expect that if they set the field and it "went through", that something was modified.  But instead, it didnt modify anything, and was just a "silent error" of sorts.

The throw at least is something the user can detect/handle, vs the silent error just leads to more of an undetectable paranoia, imo

Mutagen.Bethesda.Skyrim/Records/Common Subrecords/ConditionData.cs
Noggog: >  think it might be good if we didn't implement the properties explicitly in ConditionData 

agreed!

Mutagen.Bethesda.Skyrim/Records/Common Subrecords/ConditionData.cs
```diff
@@ -1,9 +1,39 @@
...
+
+        value = conditionData.Parameter1;
+        return value is not null;
+    }
+
+    public static bool TrySetParameter1(this IConditionData conditionData, object? value)
+    {
+        try
+        {
+            conditionData.Parameter1 = value;
```
Noggog, line 27: I think ideally it wouldn't need to throw and catch to detect a false (exceptions are a decent overhead).  In theory we could do the same check we do in the parameter set, and just return false
`if (value is not IFormLinkOrIndex<IKeywordGetter>) return false;`
Would be a hyperoptimization.. pretty low prio

Noggog (approved):

Mutagen.Bethesda.Skyrim/Records/Common Subrecords/ConditionData.cs
Elscrux: That would need to be implemented for every concrete condition data though right?

Noggog: ah true.  Probably fine for now

PR #583, opened 2024-12-30
Env asset provider

Noggog:
@Elscrux mainly wanted to make sure this API change didn't affect you negatively.   The asset provider suite has had the custom ordering parameter removed, in favor of just leaning on a dependency of the load order provider.   Felt a lot cleaner, and helped make it easier to expose an IAssetProvider through the game environment


Elscrux (approved):

Elscrux:
That should all be good. Thanks for making the AssetProvider available in the GameEnvironment. That's great to have!

PR #587, opened 2025-01-02
Add missing unused map marker from Skyrim to MapMarkerType enum

ProfessionalDragonslayer:
The MarkerType enum is missing the Castle Karstaag map marker from Skyrim. It's unused in the base game but many mods make use of it and the other unused markers (e.g. Altar, City and Smelter) are already included. I suspect the reason that it wasn't included thus far is that its id comes after the last used marker.


Noggog (approved):
Thanks for the update!

PR #596, opened 2025-02-14
Add Duplicate Mod extension method

Elscrux:
In the future the file system should be updated to be in memory.

### DIFF
--- a/Mutagen.Bethesda.Core/Plugins/Extensions/IModExt.cs
+++ b/Mutagen.Bethesda.Core/Plugins/Extensions/IModExt.cs
@@ -94,7 +94,7 @@ public static class IModExt
         return MasterStyle.Full;
     }
 
-    public static IMod Duplicate(this IModGetter mod, ModKey newModKey, BinaryWriteParameters? writeParameters = null, BinaryReadParameters? readParameters = null)
+    public static IMod Duplicate(this IModGetter mod, ModKey newModKey)
     {
         if (mod.ModKey.Type != newModKey.Type) throw new ArgumentException("ModKey types must match");
 
@@ -105,8 +105,7 @@ public static class IModExt
         var oldModPath = new ModPath(oldModKey, fileSystem.Path.Combine(fileSystemRoot, oldModKey.FileName.String));
 
         // Write mod to file system
-        writeParameters ??= BinaryWriteParameters.Default;
-        mod.WriteToBinary(oldModPath, writeParameters with
+        mod.WriteToBinary(oldModPath, BinaryWriteParameters.Default with
         {
             FileSystem = fileSystem
         });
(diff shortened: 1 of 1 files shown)


Mutagen.Bethesda.Core/Plugins/Extensions/IModExt.cs
```diff
@@ -91,4 +93,45 @@ public static MasterStyle GetMasterStyle(this IModFlagsGetter mod)
         if (medium) return MasterStyle.Medium;
         return MasterStyle.Full;
     }
+
+    public static IMod Duplicate(this IModGetter mod, ModKey newModKey, BinaryWriteParameters? writeParameters = null, BinaryReadParameters? readParameters = null)
+    {
+        if (mod.ModKey.Type != newModKey.Type) throw new ArgumentException("ModKey types must match");
```
Noggog, line 106: Ah, good add here.  I wasn't thinking about the possibility of someone trying to dup into an ESL, lol.

Mutagen.Bethesda.Core/Plugins/Extensions/IModExt.cs
```diff
@@ -91,4 +93,45 @@ public static MasterStyle GetMasterStyle(this IModFlagsGetter mod)
         if (medium) return MasterStyle.Medium;
         return MasterStyle.Full;
     }
+
+    public static IMod Duplicate(this IModGetter mod, ModKey newModKey, BinaryWriteParameters? writeParameters = null, BinaryReadParameters? readParameters = null)
```
Noggog: Are you using these write/read parameters?   I think we can maybe remove these for now and internally pick the best settings needed for getting a new mod 1:1.  Their presence here is leaking a bit of the implementation details (eg, the fact that we're using on-disk writes to do the dup)

Mutagen.Bethesda.Core/Plugins/Extensions/IModExt.cs
```diff
@@ -91,4 +93,45 @@ public static MasterStyle GetMasterStyle(this IModFlagsGetter mod)
...
+    {
+        if (mod.ModKey.Type != newModKey.Type) throw new ArgumentException("ModKey types must match");
+
+        var oldModKey = mod.ModKey;
+        var fileSystem = new FileSystem();
+        using var tmp = TempFolder.Factory();
+        var fileSystemRoot = tmp.Dir;
+        var oldModPath = new ModPath(oldModKey, fileSystem.Path.Combine(fileSystemRoot, oldModKey.FileName.String));
+
+        // Write mod to in memory file system
```
Noggog: Either delete the comment, or remove the term `memory`, as we're not actually writing to memory anymore

Noggog (approved):
Looking good!  Main thought is we can simplify the API call to not expose the sins we're doing internally, haha.

--------------

If you got time for unit tests, that's always appreciated, but not required

To dos as followups sometime:
- Design an alternative to not use on-disk systems for the dup
- Maybe offer more strongly typed alternatives per-game that handle the casting for us
```
// No cast needed
ISkyrimMod dup = skyrimMod.Duplicate(newModKey);
```
This would probably involve having the code generation put specialized duplicate function on the mod interfaces that just forward to this main one and do the appropriate casting under the hood

Mutagen.Bethesda.Core/Plugins/Extensions/IModExt.cs
Elscrux: I thought that would be good to have in case you want to modify anything. But if you think that's not needed or wanted, then I can remove it.

Noggog (approved):

PR #605, opened 2025-04-02
Sync hash code implementations for AssetLink and AssetLinkGetter

Elscrux:


Noggog:
Hmm, did you add this because the behavior was not working as expected?  If so, probably good to add a unit test.

But in general, not sure if adding the generic type -needs- to exist in the hash function?  It can, but not necessarily required (i havent done it before).

Elscrux:
> Hmm, did you add this because the behavior was not working as expected? If so, probably good to add a unit test.
> 
> But in general, not sure if adding the generic type -needs- to exist in the hash function? It can, but not necessarily required (i havent done it before).

I had AssetLink<T> and AssetLinkGetter<T> and tried to compare them (I had a Dictionary that saved the one and retrieved the other). Not sure we need a unit test for this though? It should just be the same. Maybe we can just remove the equality methods from AssetLink<T> because it's inheriting from AssetLinkGetter<T>?
I don't remember why I added the generic type, but I did at some point in the part, probably to fix something? But it might also just not be necessary to be fair. It depends if there can be two identical file paths with different asset types.

Noggog:
We can just mimic FormLink for the most part.  Mostly the same mechanics, just string instead of FormKey as the "key".

I think generally you want hashcode to be less specific rather than too specific, if it can't be "just right".  For generics it becomes harder because of inheritance.   Consider:

`FormLink<INpc>` vs `FormLink<INpcGetter>`, and let's say we want these to be considered equivalent.

Well, the equals can start to look like:
```cs
if (rhs is IFormLinkGetter<T> otherLink)
{
   return this.FormKey == otherLink.FormKey;
}
```
But note that this equality will only work one way.  `FormLink<INpc>.Equals(FormLink<INpcGetter>)` will return false, because `INpcGetter` does not implement `INpc`.  Only the reverse is true.  hence the FormLink equality does some extra stuff to check the reverse is true and yada yada

Well it gets harder for GetHashCode.  You just need to come up with -a- number.   What number do you come up with that can represent INpc OR INpcGetter?   You can't.

So rather, you just don't consider that in the GetHashCode at all.  Hashing is just a way to filter out as many obvious known non-matches as possible...   but it doesn't have to filter them all out (and in fact it can't ever do so for sure).   Just filtering on FormKey means that it'll only "collide" and potentially have a few matches if two links had the same formkey but different generics.   That's fine... each lookup starts with a GetHashCode but always ends with an Equals to confirm.. so it'll check each and find the one with a matching generic.   The hashcode itself just needed to narrow it well enough that the followup equality checks were quick.

TLDR:  Representing generics /w inheritance considerations is hard in hashcodes, imo.  And they dont need to do that job if it's too hard, so probably not worth.  

If you got something you know is failing and needs improvement, then we can take a closer look.. but again probably would want to add a unit test to sus out the failing logics edge case while we're here.

-----

I would almost say this PR should just reverse.  I'm noticing the AssetLink's code already has a 
```cs
HashCode.Combine(
 typeof(TAssetType).GetHashCode(),
 _DataRelativePath.GetHashCode());
```

maybe we actually remove that type instead of adding it to AssetLinkGetter?

-----

Last, I dont think we have any yet, but if we ever inherited two asset type classes, we'd want to swap to more advanced equality checks ala FormLink

Elscrux:
Not sure if the comparison with `FormLink<INpc>` vs `FormLink<INpcGetter>` is quite correct.
Here it is rather ``FormLinkGetter<INpc>` vs `FormLink<INpc>`. We don't have specific setter or getter types for asset types.
So it's a question if we want to allow `AssetLink<SkyrimModelAssetType>` to be able to be the equal of `AssetLink<SkyrimTextureAssetType>` or so when they both have the same path, the same FormKey in the analogy.
The form link comparison would be if we should allow `FormLink<INpc>` to equal `FormLink<IWeapon>` or so.

Noggog:
Yeah, for sure.   We dont have SkyrimModelGetter vs SkyrimModel.  It would work right now with our current "flat" asset type structure as there's none of that complexity, like you're saying. 

But the same concerns would apply if we had any inheritance added to the suite at all in the future.   Like, if anyone was to ever do `public class SkyrimSimpleModel : SkyrimModel` or something like that.   Having the type included in the GetHashCode is fragile in a subtle way that would break silently if someone were to add anything like that.   

No real right answer, cuz we could definitely just enforce NO inheritance in asset types as well and be just fine.    But the upside of including the type in the hashcode is minimal, so not worth the subtle bug tradeoff, imo?

Elscrux:
Yeah, I think the main concern is not if we can include the type technically yeah.
The question is if we should think that equality of an asset link is just based on the path. Are there any options where we could potentially have the same path for two different asset types? Currently I don't think we have overlapping data types like that (well I think we already have separate music and sfx asset types which both have .wav files but they still different base directories).
If we see equality only based on the path, we have the design decision that asset types cannot collide in any way. We can do that but we need to document it probably and maybe even add a test or so that makes sure we don't do this.

Noggog:
> The question is if we should think that equality of an asset link is just based on the path

I don't think that we'd be forced to define the equality that way.

HashCodes just need to avoid false negatives, where two equal things somehow dont return the same hashcode  (uh oh).  They're allowed to return a false positive, where two not-equal things return the same hashcode. 

So, I think it's fine if the type isn't included in the hash code and would lead to slightly more collisions.    We can still define that asset links are only equal if both the path AND type make sense

Elscrux:
Yeah just for HashCodes it's probably fine! Sorry, I should make clear that I'm talking about general equality that we'd want to make more consistent. There we currently include the type in the check, seeing as we use equals functions like this:
```cs
    public override bool Equals(object? obj)
    {
        if (ReferenceEquals(null, obj)) return false;
        if (ReferenceEquals(this, obj)) return true;
        if (obj is not IAssetLinkGetter<TAssetType> rhs) return false;

        return DataRelativePath.Equals(rhs.DataRelativePath);
    }
```

Noggog:
No worries.  Also about to crash for sleep, so not reading between lines too well haha.

On the original qusetion, i think it's better to view asset link equality as both path and type.  Makes most sense to me?

In that case, the above snippet looks fine enough for now.  It'd need to be upgraded to FormLink's style if we added asset types that inherited from one another

Elscrux:
Cool I've changed it to remove the type from the hash code and we keep the type check in the equality check.

Noggog (approved):

PR #608, opened 2025-04-19
Finalize Font Implementation

Elscrux:
Finalize implementation of font logic. Just needed a Module for DI and there were some leftover debug messages.

### DIFF
--- a/Mutagen.Bethesda.Core/Fonts/FontProvider.cs
+++ b/Mutagen.Bethesda.Core/Fonts/FontProvider.cs
@@ -1,6 +1,8 @@
 ﻿using Mutagen.Bethesda.Assets;
+using Mutagen.Bethesda.Fonts.DI;
 using Mutagen.Bethesda.Strings;
-namespace Mutagen.Bethesda.Fonts.DI;
+
+namespace Mutagen.Bethesda.Fonts;
 
 public interface IFontProvider {
 	/// <summary>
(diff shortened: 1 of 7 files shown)


Mutagen.Bethesda.Core/Fonts/FontProvider.cs
```diff
@@ -1,6 +1,8 @@
 ﻿using Mutagen.Bethesda.Assets;
+using Mutagen.Bethesda.Fonts.DI;
 using Mutagen.Bethesda.Strings;
-namespace Mutagen.Bethesda.Fonts.DI;
+
+namespace Mutagen.Bethesda.Fonts;
```
Noggog, line 5: Not this line, but further down... might be good to move the concrete class FontProvider to DI? 

The current pattern we've been following so far was that if someone wanted to get access to this functionality without using DI, their entry point would often be a static class, like `Fonts.GetProvider(Language)` or whatever, which under the hood would make the DI components and return them for the user.   `LoadOrder` static class being an example of this non-DI entrypoint style.

I think the FontProvider as written feels more like it should go on the DI side of things?

Noggog (approved):
Looks good!  Only comment is maybe moving FontProvider back into DI land

Mutagen.Bethesda.Core/Fonts/FontProvider.cs
Elscrux: Yeah I envisioned FontProvider to be a class used in DI, but seeing as you need to provide the language to create an instance of the class it needed to be set up differently which is why I made the factory class. A static class wouldn't work sadly I think as you still need access to the IGetFontConfig implementation to read the config which you wouldn't have in a static class.

PR #610, opened 2025-05-03
Resolved sounds dialog responses

Elscrux:
I noticed I never actually added assets for the responses themselves in case you just want assets from a specific response.


Elscrux:
Though I did notice for this that adding it to the IDialogResponsesGetter actually creates some issues as now it would report the assets for the DialogTopic and the nested DialogResponses which are the same. I'll remove the DialogTopic asset enumeration directly in DialogTopic because of that and handle it just through the nested enumeration of DialogResponses.

Noggog (approved):

PR #618, opened 2025-05-25
Expanded IPlaced

Elscrux:
- Expand IPlaced interface with all common members
- Remove ILocationRecordGetter link interface


Noggog (commented):
Looks generally good and fine.

Can we confirm that xEdit is wrong in their definitions here?
https://github.com/TES5Edit/TES5Edit/blob/c64e504d06bb59502be8a0a52b6a7fcf184c3ebd/Core/wbDefinitionsTES5.pas#L2658

Elscrux:
> Looks generally good and fine.
> 
> Can we confirm that xEdit is wrong in their definitions here? https://github.com/TES5Edit/TES5Edit/blob/c64e504d06bb59502be8a0a52b6a7fcf184c3ebd/Core/wbDefinitionsTES5.pas#L2658

I have checked this one yeah. The only purpose here can be that a location is referenced, this is about tying it to that location in case you have a record that overwritten by another mod to make sure it's in the right location. There is one reference in Skyrim where a location ref type is used but that seems like an error in Skyrim. I searched the xEdit Server for more info on this and found no reference of any use for a location ref type. It wouldn't really make sense anyway, so I though I'd just get rid of it.

Noggog (approved):

PR #626, opened 2025-06-04
Voice type lookup tweaks

Elscrux:


Elscrux:
Things should be ready now. I updated it to cover some more edge cases that I missed previously.
Fortunately all this previous work on this was very helpful to create a very helpful analyzer

Mutagen.Bethesda.Skyrim/Records/Assets/VoiceType/VoiceTypeAssetLookup.cs
```diff
@@ -10,7 +11,7 @@ public class VoiceTypeAssetLookup : IAssetCacheComponent
     private ILinkCache _formLinkCache = null!;
 
     //Databases
-    private readonly Dictionary<ModKey, VoiceContainer> _defaultSpeakerVoices = new();
+    private readonly ConcurrentDictionary<ModKey, VoiceContainer> _defaultSpeakerVoices = new();
```
Noggog: Let's perhaps be a bit clearer on what is part of the `Prep` phase, and thus completely threadsafe, vs the things that are lazy and on-demand and thus aren't threadsafe.   We have Databases vs Caches, but im not sure what the differences are at the moment.

In those cases, I think it's better to just do a 
`object _lock = new object()` and do a normal lock block around accesses of those operations?

Noggog (changes requested):
Up to you, but I think this component has gotten complex and specialized enough to warrant unit tests to flex the expected behavior.   I think at this point if i were to ever try to maintain this component, I'd be a bit of a bull in a china shop, likely breaking features I didn't know were happening?  Tests might clarify expected behaviors, such as whatever improvement this PR is adding

At the least, though, let's do the improvements to the locking patterns, I'd say?

Elscrux:
> Up to you, but I think this component has gotten complex and specialized enough to warrant unit tests to flex the expected behavior. I think at this point if i were to ever try to maintain this component, I'd be a bit of a bull in a china shop, likely breaking features I didn't know were happening? Tests might clarify expected behaviors, such as whatever improvement this PR is adding

It's definitely not simple logic. Sadly this functionality requires lots of complex case handling. We already have a number of tests that ensure that the changes in this PR are correct though.

Noggog:
Gotcha.  So you'd say this PR doesn't change any behavior, it's just internal implementation improvements?   If it's fixing something, then that means the behavior is changing and we could write a test that shows it's now behaving as it should in that regard

Mutagen.Bethesda.Skyrim/Records/Assets/VoiceType/VoiceTypeAssetLookup.cs
Elscrux: Yeah seems like that distinction between databases (I think that's supposed to be static stuff we fully parse at the start) vs caches (stuff you cache when you need it).

Why would you use locks?

Elscrux:
> Gotcha. So you'd say this PR doesn't change any behavior, it's just internal implementation improvements? If it's fixing something, then that means the behavior is changing and we could write a test that shows it's now behaving as it should in that regard

There are some fixes in there but those cases are so complex to set up that I didn't really feel like to set up a test case with 20 setup steps for it, sorry :/ I know it'd be better rationally :D

Mutagen.Bethesda.Skyrim/Records/Assets/VoiceType/VoiceTypeAssetLookup.cs
```diff
@@ -168,40 +203,14 @@ public void Prep(IAssetLinkCache linkCache)
         else
         {
             questVoices = GetVoices(quest, topic.FormKey.ModKey);
-            _questCache.Add(quest.FormKey, questVoices);
+            _questCache.TryAdd(quest.FormKey, questVoices);
```
Noggog: Can we reuse the `GetQuestVoices` call here?

Mutagen.Bethesda.Skyrim/Records/Assets/VoiceType/VoiceTypeAssetLookup.cs
Noggog: It's tradeoffs and designs.  

ConcurrentDictionaries can be nice, but in some scenarios it'll go like this:
- User requests operation A on Thread X
- User requests operation A on Thread Y
- X sees that A isn't in the dict, starts work making A
- Y sees that A isn't in the dict, starts work making A
- X finishes work A inserts it
- Y finishes work A, replaces the same result

In that scenario, you could have 10 threads doing the same work A, compared to a lock would block thread Y from accessing the dictionary at all until X finished the work on it.  

That's nicer for not redoing the work of A multiple times, but still, Y is blocked doing nothing until then.

But, I guess that's worse behavior if Y was doing work on B, it would be blocked until A finished.   So maybe the ConcurrentDictionary is nice to have in the situations that different threads are doing different work.

---------

Another reason for using normal locks is if there simplifies interactions between multiple dictionaries that are intertwined.  Removes the edge cases where one has been updated but the other hasn't yet, if that's a problem.   I guess that's not actually happening too much.  Seems more like several self sufficient dictionaries rather than two dictionaries that need to be "in sync".

------------

I guess keeping ConcurrentDictionaries are fine here.   Maybe just organize the variables and naming to be clearer on which are these on-demand caches

Mutagen.Bethesda.Skyrim/Records/Assets/VoiceType/VoiceTypeAssetLookup.cs
Elscrux: I think so - good point

Noggog:
> > Gotcha. So you'd say this PR doesn't change any behavior, it's just internal implementation improvements? If it's fixing something, then that means the behavior is changing and we could write a test that shows it's now behaving as it should in that regard
> 
> There are some fixes in there but those cases are so complex to set up that I didn't really feel like to set up a test case with 20 setup steps for it, sorry :/ I know it'd be better rationally :D

haha all good.   Catch 22 a bit, as the complexity is what makes the tests more interesting (as it is complex and thus easy to get wrong).  But yeah, lots of work.  Up to you!

Mutagen.Bethesda.Skyrim/Records/Assets/VoiceType/VoiceTypeAssetLookup.cs
Elscrux: Oh, that's a good point, hadn't thought about redundant computations!

I've switched it out with locks

Noggog (approved):

PR #627, opened 2025-06-11
Initial IReferencedByCache work

Noggog:
Initial interface design for `IReferencedByCache`.  It's backed by a class that should do the job, but won't be as optimized as it can be, as per the ToDo comment.  We'll mix in the optimizations in followup PRs, I think.


Mutagen.Bethesda.Core/Plugins/Cache/IReferencedByCache.cs
```diff
@@ -0,0 +1,17 @@
+﻿using Mutagen.Bethesda.Plugins.Records;
+
+namespace Mutagen.Bethesda.Plugins.Cache;
+
+public interface IReferencedByCache
+{
+    IReadOnlyCollection<IFormLinkGetter<TReferencedBy>> GetReferencedBy<TReferencedBy>(
```
Elscrux: We'll also need a variant that is returning any `IFormLinkIdentifier`, not just certain typed ones.

Mutagen.Bethesda.Core/Plugins/Cache/Internals/Implementations/ImmutableLoadOrderReferencedByCache.cs
```diff
@@ -0,0 +1,98 @@
...
+        
+        lock (_cache)
+        {
+            if (_cache.TryGetValue(key, out var cached)) return cached.FormKeys ?? [];
+        }
+        
+        HashSet<FormKey> result = new();
+        foreach (var record in _linkCache.PriorityOrder.WinningOverrides<IMajorRecordGetter>()) 
+        {
+            if (record.EnumerateFormLinks().Any(reference => reference.FormKey == formKey))
```
Elscrux: Here would could already cache all references we find when we enumerate them anyway.
So you don't have to go through all placed objects again to find IronDagger next to IronSword that you previously searched the references for.
This would make the first call slower though, but for more calls in the future that would be better. Maybe we need some tests to see what makes most sense in terms of speed (initial speed, and speed over time).

Mutagen.Bethesda.Core/Plugins/Cache/Internals/Implementations/ImmutableLoadOrderReferencedByCache.cs
```diff
@@ -0,0 +1,98 @@
...
+
+        lock (_cache)
+        {
+            if (_cache.TryGetValue(key, out var cached))
+            {
+                if (cached.FormLinks is IReadOnlyCollection<IFormLinkGetter<TReferencedBy>> links)
+                {
+                    return links;
+                }
+                return [];
```
Elscrux: I guess this is something we should never reach? Might make sense to add an exception here?

Mutagen.Bethesda.Core/Plugins/Cache/IReferencedByCache.cs
```diff
@@ -0,0 +1,17 @@
...
+        IFormLinkIdentifier identifier)
+        where TReferencedBy : class, IMajorRecordGetter;
+
+    IReadOnlyCollection<IFormLinkGetter<TReferencedBy>> GetReferencedBy<TReferencedBy>(
```
Elscrux: General thought about the name ReferencedBy. We're talking record that we want to get the references of.

Currently it reads like "get me all that is referenced by the record I pass in". But it's the other way around and what you get is "all the references to the record I pass in". I see the ReferencedBy thing being a nod to how xEdit calls its menu when you have a record selected, but for this it doesn't quite feel right.

Not quite sure what works best, here are some ideas:
- GetReferencesTo
- GetReferences
- GetLinksTo
- GetLinks
- GetUsages

+ variants with explicit mentions of "Record". References/links might also be asset links or similar.

Worth noting is also that "reference" is not really a term in Mutagen, and it's rather using the term Link to describe things relating to other things. Reference might be good as a different term to make it clear that it's sort of the backward links, but that might also just be confusing for people who don't understand what is meant here and now we're talking about links and about references which are similar terms that are doing very different things.

Maybe we should even go more into the direction of what the `ILinkCache` api looks like right now and call it something like.
- ResolveLinksTo
- ResolveLinks

With the resolve, you also get the IntelliSense for this so it's easily discoverable for devs. Just needs to be clearly a different thing.

Noggog: I view the term `Resolve` as meaning the act of getting an actual record object from a FormLink.   In this case, we're just getting a list of links from a link, so there's no Resolving going on

But i have renamed the bits.  Let me know how they look now!

Mutagen.Bethesda.Core/Plugins/Cache/ILinkUsageCache.cs
```diff
@@ -0,0 +1,23 @@
+﻿using Mutagen.Bethesda.Plugins.Records;
+
+namespace Mutagen.Bethesda.Plugins.Cache;
+
+public interface ILinkUsageCache
+{
+    IReadOnlyCollection<IFormLinkGetter<TReferencedBy>> GetUsagesOf<TReferencedBy>(
```
Elscrux: I guess we should also rename TReferencedBy to TUsage now

Mutagen.Bethesda.Core/Plugins/Cache/ILinkUsageCache.cs
```diff
@@ -0,0 +1,23 @@
...
+        IFormLinkIdentifier identifier)
+        where TReferencedBy : class, IMajorRecordGetter;
+    
+    IReadOnlyCollection<FormKey> GetUsagesOf(
```
Elscrux: Can we make all these return IFormLinkIdentifiers instead of FormKeys? It would be pretty useful to have the type information in there.

Mutagen.Bethesda.Core/Plugins/Cache/Internals/Implementations/ImmutableLoadOrderLinkUsageCache.cs
```diff
@@ -0,0 +1,168 @@
...
+        var key = new CacheKey(identifier.FormKey, 
+            ReferencedType: identifier.Type,
+            ReferencedByType: typeof(IMajorRecordGetter));
+        
+        lock (_cache)
+        {
+            if (_cache.TryGetValue(key, out var cached)) return cached.FormKeys;
+        }
+        
+        var genMethod = MethodInfo.MakeGenericMethod(typeof(IMajorRecordGetter));
```
Elscrux: Why does this need to be a reflection method call? Couldn't it be `GetReferencedByGeneric<IMajorRecordGetter>(identifier)`?

Mutagen.Bethesda.Core/Plugins/Cache/Internals/Implementations/ImmutableLoadOrderLinkUsageCache.cs
```diff
@@ -0,0 +1,168 @@
...
+                }
+            }
+            return _untypedReferencesCache;
+        }
+    }
+
+    public IReadOnlyCollection<FormKey> GetUsagesOf(FormKey formKey)
+    {
+        var key = new CacheKey(formKey, 
+            ReferencedType: typeof(IMajorRecordGetter),
```
Elscrux: In cases like this I think we shouldn't split up the key per type we passed in here and disregard work that we might have done on the specifically typed form link with the given form key before?

Mutagen.Bethesda.Core/Plugins/Cache/Internals/Implementations/ImmutableLoadOrderLinkUsageCache.cs
```diff
@@ -0,0 +1,168 @@
...
+        MethodInfo = typeof(ImmutableLoadOrderLinkUsageCache).GetMethod(nameof(GetReferencedByGeneric), BindingFlags.NonPublic | BindingFlags.Instance)!;
+    }
+    
+    private record struct CacheKey(
+        FormKey FormKey,
+        Type ReferencedType,
+        Type ReferencedByType);
+    
+    private record CacheItem(
+        IReadOnlyCollection<FormKey> FormKeys,
```
Elscrux: Also related to what I mentioned that we'd wanted to get IFormLinkIdentifiers as a result for usage calls - do we really need this to be split up into form keys and form links?

Mutagen.Bethesda.Core/Plugins/Cache/Internals/Implementations/ImmutableLoadOrderLinkUsageCache.cs
```diff
@@ -0,0 +1,168 @@
...
+            }
+            return _untypedReferencesCache;
+        }
+    }
+
+    public IReadOnlyCollection<FormKey> GetUsagesOf(FormKey formKey)
+    {
+        var key = new CacheKey(formKey, 
+            ReferencedType: typeof(IMajorRecordGetter),
+            ReferencedByType: typeof(IMajorRecordGetter));
```
Elscrux: Instead of having separate keys for IMajorRecordGetter usages and I think it might be better to just have form keys as key, and then per form key possible types of usages. If we looked at types A and B we can save that there and return the cached results for it. If we want to look at all types we can best already skip A and B and look at the record of the types and cache them.

Elscrux (commented):
I like the naming, I think that should make it pretty clear what this is about! Maybe change the remaining internal mentions of Reference / ReferencedBy too for consistency?
As far as the caching is concerned I'm a bit unsure if this is the best approach right now as we split things a lot between untyped and typed approaches where a lot of the same work is done multiple times potentially. I've explained it more down below, but I think this kind of caching structure could make more sense.
Naive structure to give you an idea what structure I mean:
```cs
Dictionary<FormKey, Dictionary<Type, HashSet<IFormLinkIdentifier>>
```
That makes it possible to get all currently cached results for a given FormKey, know what work can be skipped and provides a common place and format to save things in.

Mutagen.Bethesda.Core/Plugins/Cache/ILinkUsageCache.cs
Noggog: If we want this sort of functionality, this is where I'd like to instead return a structure like `UsageResults`.   The return value as of now makes it easy to then take the usage results and check `Contains(someFormKey)` to find out if something references something else.   But if it was a `IReadOnlyCollection<IFormLinkIdentifier>` then the call might fail or not line up until the user make a downstream `.Select(x => x.FormKey).ToHashSet()` to then do the Contains call on, etc.

By offering a `UsageResults` struct, we can return a "smarter" object that is able to do various typical things off the same cached results.

Mutagen.Bethesda.Core/Plugins/Cache/Internals/Implementations/ImmutableLoadOrderLinkUsageCache.cs
Noggog: Isn't fancier always better??!  Fixed

Mutagen.Bethesda.Core/Plugins/Cache/Internals/Implementations/ImmutableLoadOrderLinkUsageCache.cs
Noggog: I'd say the ratio of time saved compared to complexity added is low.   At the moment, we just are calling EnumerateMajorRecords(), which doesn't have a way of enumerating all record types -except- A and B.  We'd have to come up with some sort of mechanism to do the skipping.

We could definitely add the complexity to extract the potential savings, but maybe as a followup.  I think for most scenarios, it'll be not applicable?

Noggog:
I think mainly just need to figure out what we want to return and how.

You'd like FormLinkIdentifiers to be able to sense the types.  
I think people will also want to do easy Contains calls to see if a specific FormKey is referencing the thing.
I'd like to offer both without requiring the user to construct a new collection from the results

Ill think on it more tomorrow

Elscrux:
> I think mainly just need to figure out what we want to return and how.
> 
> You'd like FormLinkIdentifiers to be able to sense the types. I think people will also want to do easy Contains calls to see if a specific FormKey is referencing the thing. I'd like to offer both without requiring the user to construct a new collection from the results
> 
> Ill think on it more tomorrow

I think a `UsageResults` or so kind of structure like you described could be fine to include if needed. I'd be basically just something to make the use cases easier that you're describing, and possibly also something for future entry points + maybe extension methods that could be defined on it.
```cs
var usageInfo = GetUsagesOf(myRecord);

// Possible scenarios
usageInfo.Contains<IArmorGetter>(formKey);
usageInfo.Contains(formLink);
usageInfo.Contains(formKey);

foreach (var link in usageInfo.EnumerateLinks()) { ... }

foreach (var link in usageInfo.EnumerateLinks<IArmorGetter>()) { ... }
```

For when we get to the internals, I still have my concerns. I think it'd be good to have some alternatives to benchmark against each other in various scenarios with repeated/redundant calls and so on - see the alternatives I sent via Discord for that btw.

Noggog:
Aight, finally got to adding the usage results object.  Had to do some general equality/contains improvements overall

Elscrux:
Some thoughts, not sure if we talked about this before so sorry if I'm repeating myself:
- Anything using the generic `IMajorRecordGetter` is loosing all type information so it would require the brute force Mutagen methods to do anything with them which is quite unfortunate. Specifically the user facing part here would be `UsageLinks` which is a set of `IFormLinkGetter<TScope>` atm. My record reference cache implementation just returned `IFormLinkIdentifier` which still kept the type, and just didn't have any generic parameter which works very well. We can maybe have a variant of `ILinkUsageResults` without generic parameter or so?
- Currently we don't use `ILinkUsageResults<IMajorRecordGetter> GetUsagesOf(FormKey formKey)` at all, it's also not part of of the `ILinkUsageCache` interface. Is this something we still have plans for?
- I know we only wanted to look into this later but do we have any plan how to deal with updating link caches? Because there is a lot of work that won't need any updating - at least for all the immutable mods in a load order which is usually most of them, and those are usually the big mods too. So keeping the option to save results per mod would be needed in this case.

Noggog:
> Anything using the generic IMajorRecordGetter is loosing all type information

I don't quite think so?  The generic just scopes it to -at least- be that type.  So when the API says it's returning `IFormLinkGetter<IMajorRecordGetter>`, it just means that the links are allowed to be of any type.  But this doesn't mean that the types the links actually are are discarded.

For example, if i had a `List<IAnimal>`, i could put classes `Cat` and `Dog` in there.   Their concrete class types remain, even though the returned list is of the umbrella type `IAnimal`.

Part of this PR upgrades internals so that the FormLinks that are made have the appropriate granular type.   This means that in this example:

```cs
using var env = GameEnvironment.Typical.Skyrim(SkyrimRelease.SkyrimSE);
INpcGetter someNpc = Skyrim.Npc.Fridrika.Resolve(env.LinkCache);
IRaceGetter race = someNpc.Race.Resolve(env.LinkCache);

var usageCache = new ImmutableLoadOrderLinkUsageCache(env.LinkCache);
var usages = usageCache.GetUsagesOf(race);
Console.WriteLine($"Was race referenced by npc? {usages.Contains(someNpc)}");
foreach (var usageLink in usages.UsageLinks)
{
    Console.WriteLine($"Link had generic {usageLink.GetType().GenericTypeArguments[0]}, and listed type {usageLink.Type}");
}
```
Prints
```
Was race referenced by npc? True
Link had generic Mutagen.Bethesda.Skyrim.INpcGetter, and listed type Mutagen.Bethesda.Skyrim.INpcGetter
...
```
This is because when the usage cache finds that the npc uses the race and wants to put that in its cache, it makes a `new FormLink<INpcGetter>`

This of course "fits" underneath `IFormLinkGetter<IMajorRecordGetter>` since `INpcGetter` is a major record.   But the type of the concrete class remains, and is still `FormLink<INpcGetter>`.

TLDR:  the return value of `ILinkUsageResults<IMajorRecordGetter>` in the least scoped scenario just means that it might contain links to npcs or races or weapons or anything.  But each individual link retains its concrete type, and so would list their appropriate type, and still be optimized when resolving with them.  (or if they weren't, that'd be a bug to be fixed)


> Currently we don't use ILinkUsageResults<IMajorRecordGetter> GetUsagesOf(FormKey formKey) at all, it's also not part of of the ILinkUsageCache interface. Is this something we still have plans for?

I think it is part of the ILinkUsageCache interface?  L59 currently is
```cs
    [Obsolete("This call is not as optimized as its generic typed counterpart.  Use as a last resort.")]
    ILinkUsageResults<IMajorRecordGetter> GetUsagesOf(FormKey formKey);
```
I would recommend against people using it when they can, but it's an option


> I know we only wanted to look into this later but do we have any plan how to deal with updating link caches

Not at the moment.  I'm focused on overall API, and then on making it as fast as possible.  I think we will be pleasantly surprised at how fast it can do it once the full plan is in place and used properly.  If that doesn't pan out, then we can reassess next steps

Elscrux:
Oh, that's great to see that the type information is kept here now! Also thank you for the explanation :)

> I think it is part of the ILinkUsageCache interface? L59 currently is

Missed that! Do you think we actually need this separate logic for passing in the form key? Because `ILinkUsageResults<IMajorRecordGetter> GetUsagesOf(IFormLinkIdentifier identifier)` is doing the exact same while only ever using the `FormKey` of `IFormLinkIdentifier` and never the type information. So it's exactly the same as just passing in the `FormKey` like we do it in `ILinkUsageResults<IMajorRecordGetter> GetUsagesOf(FormKey formKey)`, just that here we have a totally separate logic that also has a separate cache which seems redundant.

Noggog:
Cool, removed the separate logic for FormKeys.  I think that was necessary initially, but with the new FormLinkInformation equality upgrades it works without

Noggog:
Aight, gonna merge this in soon, i think.   Be aware that the actual implementation is literal worst case situation as far as speed.  So hopefully will work correctly, but just be slow until the internals are upgraded

Elscrux:
> Cool, removed the separate logic for FormKeys. I think that was necessary initially, but with the new FormLinkInformation equality upgrades it works without

@Noggog What I was referring to was `ILinkUsageResults<IMajorRecordGetter> GetUsagesOf(FormKey formKey)` which is still in there btw. It seems like it does the exact same as `ILinkUsageResults<IMajorRecordGetter> GetUsagesOf(IFormLinkIdentifier identifier)` but with a separate cache that nobody else uses. So not sure if that has any reason to exist still.

Noggog:
> but with a separate cache that nobody else uses

I believe i removed the separate cache, and the logic now reroutes to use the IFormLinkIdentifier logic under the hood.

The reason it's nice to expose as a `(FormKey formKey)` is that this is for "newb" users that dont know better but "just want it to work!?!".   To lean on the FormLinkInformation in the same way would require them to do `new FormLinkInformation(theFormKey, typeof(IMajorRecordGetter))`, which is an unnecessary annoyance.   [The current code just does that under the hood for them](https://github.com/Mutagen-Modding/Mutagen/blob/f1888d3d326a9e722bdaf2b9de789f8bca4d0f04/Mutagen.Bethesda.Core/Plugins/Cache/Internals/Implementations/ImmutableLoadOrderLinkUsageCache.cs#L176), but shouldnt have that 2nd cache anymore unless i missed one

Elscrux:
> > but with a separate cache that nobody else uses
> 
> I believe i removed the separate cache, and the logic now reroutes to use the IFormLinkIdentifier logic under the hood.
> 
> The reason it's nice to expose as a `(FormKey formKey)` is that this is for "newb" users that dont know better but "just want it to work!?!". To lean on the FormLinkInformation in the same way would require them to do `new FormLinkInformation(theFormKey, typeof(IMajorRecordGetter))`, which is an unnecessary annoyance. [The current code just does that under the hood for them](https://github.com/Mutagen-Modding/Mutagen/blob/f1888d3d326a9e722bdaf2b9de789f8bca4d0f04/Mutagen.Bethesda.Core/Plugins/Cache/Internals/Implementations/ImmutableLoadOrderLinkUsageCache.cs#L176), but shouldnt have that 2nd cache anymore unless i missed one

The thing is that `ILinkUsageResults<IMajorRecordGetter> GetUsagesOf(IFormLinkIdentifier identifier)` never actually uses the type of the identifier. So it's the same as supplying FormKey. This could just be two override variants of the method joining into the same implementation.

Noggog:
Ah, good catch.  Ill add a test and make another PR to fix that

Noggog:
https://github.com/Mutagen-Modding/Mutagen/pull/632

PR #630, opened 2025-07-04
Add Effects list to Effect Record interface

Elscrux:


Noggog (changes requested):
I think most interfaces like this should either:
- Be a LinkInterface if it's something a FormLink can point to (like IPlaced)
- Be an Aspect interface, passed out automatically via the AspectInterfaceModule and friends

Those two routes maintain a lot of other mechanisms like the `AspectInterfaceMapping_Generated.cs` file which constructs lookups for things like `IMetaInterfaceMapGetter` and `IAspectInterfaceMapGetter`

Adding the interfaces ad-hoc will miss those extra registration bits  I think what you have here should work fairly well as a `RefAspect` definition, similar to how IModeled is set up in `AspectInterfaceModule.cs`?

Elscrux:
> I think most interfaces like this should either:
> 
>     * Be a LinkInterface if it's something a FormLink can point to (like IPlaced)
> 
>     * Be an Aspect interface, passed out automatically via the AspectInterfaceModule and friends
> 
> 
> Those two routes maintain a lot of other mechanisms like the `AspectInterfaceMapping_Generated.cs` file which constructs lookups for things like `IMetaInterfaceMapGetter` and `IAspectInterfaceMapGetter`
> 
> Adding the interfaces ad-hoc will miss those extra registration bits I think what you have here should work fairly well as a `RefAspect` definition, similar to how IModeled is set up in `AspectInterfaceModule.cs`?

My motivation for this was to be able to add an extension function that works on Spell, Scroll, Ingestible, Ingredient and ObjectEffect at the same time. So the idea is definitely more for an aspect interface, and not a link interface (here it seems like only Spell and ObjectEffect are actually originally references by IEffectRecord anyway. So I think I need to create a new aspect interface for that. `IHasEffects` maybe?

Btw I think my reference for this implementation was `IPlaced` which was also set up this way with members on the interface even though it's a link interface. Is that a problem? I always regarded IPlaced as sort the base interface for the placed records, and not specifically an aspect or link interface, or both at the same time.

Elscrux:
@Noggog I've replaced the work so far with the new aspect interface IHasEffects

Mutagen.Bethesda.Starfield/Records/Major Records/LightGobo_Generated.cs
```diff
@@ -2026,7 +2026,7 @@ public static void WriteEmbedded(
             FloatBinaryTranslation<MutagenFrame, MutagenWriter>.Instance.Write(
                 writer: writer,
                 item: item.StaticRotation,
-                divisor: 57.295799255371094f,
+                divisor: 57,295799255371094f,
```
Noggog: Hmm, this seems unintended side effect.   Probably a bug in the generator code that isn't enforcing culture correctly.

Can you toy around with the FloatType.cs DivisorString to get it to print `.` instead of `,`?  Also relates to `FloatBinaryTranslationGeneration.cs`

Noggog (changes requested):
Main part of PR looking good

Mutagen.Bethesda.Starfield/Records/Major Records/LightGobo_Generated.cs
Elscrux: Oh, yeah that's an annoyance I've had since forever. Would be great if that could be fixed.

Elscrux: @Noggog Last time I looked into it, it seemed like this problem is in the Loqui library.

Noggog: Could also be a few things to fix in Loqui itself, but I think this specific one is in the Mutagen repo's `FloatType`, if i had to guess

Elscrux: Ok I fixed it, although I got a few strange generation results after. Floats randomly changed to be a bit lower. Could you take a look? I don't see a semantic difference in the code apart from the special culture info string I used to get rid of the comma. So I guess it might just be some weird side effect from that?

Mutagen.Bethesda.Starfield/Records/Major Records/OrbitedData_Generated.cs
```diff
@@ -1160,7 +1160,7 @@ public static void WriteEmbedded(
                 writer: writer,
                 item: item.MassComparedToEarth,
                 divisor: null,
-                multiplier: 5.972E+24f);
+                multiplier: 5.972E+27f);
```
Noggog: That is odd... when i check the branch out and regenerate, i get the original+24

ConsoleApp2/ConsoleApp2.csproj
```diff
@@ -0,0 +1,14 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+    <PropertyGroup>
```
Noggog: Leftover from some testing?

Mutagen.Bethesda.Starfield/Records/Major Records/OrbitedData_Generated.cs
Noggog: You can see the value is set fairly straightfoward/direct in Planet.xml's OrbitedData xml area
`<Float name="MassComparedToEarth" divisor="5.972E24" />`
So not sure how yours is somehow converting that to a E27
Maybe put a breakpoint in FloatType's Load call, with a conditional of `Name == "MassComparedToEarth"` and step through to see if/when it somehow gets to the E27 instead of the value expressed by the string `5.972E24`?

Noggog (changes requested):

Mutagen.Bethesda.Starfield/Records/Major Records/OrbitedData_Generated.cs
Noggog: Maybe it's the 
`Multiplier = node.GetAttribute("multiplier", 1d);`
Where the logic isn't parsing it with invariant culture?  Still, odd that it would be the E24 that gets misapplied.

Maybe try 
```
if (node.TryGetAttributeString("multiplier", out var multStr))
{
    Multiplier = double.Parse(multStr, CultureInfo.InvariantCulture);
}
```

Elscrux: Seems like `Convert.ChangeType((object) strVal, type)` inside `TryGetAttribute<P>(this XElement node, string str,[MaybeNullWhen(false)] out P val, bool throwException = false)` is doing this for me.

I can reproduce that this `Convert.ChangeType("5.972E24", typeof(double));` outputs this: "5,972E+27". Some for type float. This works correctly `Convert.ChangeType("5.972E24", typeof(double), CultureInfo.InvariantCulture);`. Although I have no idea why this would be a culture thing but well. Seeing as this is part of the `XmlExt` class in your Noggog library, could you take a look please?

Noggog: Yeah, let me update that on my end

Noggog:
Aight, if you pull latest dev, it should be using CSharpExt 2.73.1 with updated culture settings

Elscrux:
> Aight, if you pull latest dev, it should be using CSharpExt 2.73.1 with updated culture settings

It seems to work well when I supply `CultureInfo.InvariantCulture` to `TryGetAttribute` but an another potential problem seems to be `GetAttribute` which would need an override for the culture as well.

Noggog:
Aight, `Mutagen` `Dev` branch is using v 2.73.2 of CSharpExt.  Should be able to pass in invariant culture parameters to the GetAttribute calls now

Elscrux:
@Noggog Thanks! I added it, and I think things should be good now. Great that we were able to figure this out, will make things a lot more comfortable to work with in the future :)

Noggog (approved):

PR #632, opened 2025-07-12
Improvements for typing for FormKeys and Identifiers in Usage Cache

Noggog:
Added "miss" edge cases to the unit tests, and made some typing improvements so they're addressed


Elscrux (approved):
Looks good! :)

Some minor comments
`_untypedReferencesCacheLock`, `_untypedReferencesCache`, and `_emptySet` could be removed as they're not used anymore and `CacheItem` and `Result` could both be sealed.

PR #635, opened 2025-08-10
Fix GetSpeakers edge cases for Shared Infos

Elscrux:
Actually return speakers based on the conditions of the lines in GetSpeakers
- Don't return no voices when parsing responses that are using shared infos
- Don't return only voices limited to all usages when parsing shared infos


Elscrux:
@Noggog added some more fixed, it should be good to go now! :)

Noggog (approved):

PR #639, opened 2025-09-03
Fix Linked Reference Package Target

Elscrux:
Change package target reference to linked reference and the record to keyword.
See https://github.com/TES5Edit/TES5Edit/blob/929cfc5da916c5fbed3e95f73619576de1ba8ad8/Core/wbDefinitionsTES5.pas#L8922


Mutagen.Bethesda.Fallout4/Records/Major Records/Package.xml
```diff
@@ -139,9 +139,9 @@
       <Enum name="Type" enumName="TargetObjectType" />
     </Fields>
   </Object>
-  <Object name="PackageTargetReference" baseClass="APackageTarget" objType="Subrecord">
+  <Object name="PackageTargetLinkedReference" baseClass="APackageTarget" objType="Subrecord">
     <Fields>
-      <FormLink name="Reference" refName="Fallout4MajorRecord" />
+      <FormLink name="Keyword" refName="Keyword" />
```
Noggog, line 144: How are you sure this is keyword only?   In your xEdit linked code, it's `wbFormID('Reference')`, which suggests it might be many things

Mutagen.Bethesda.Fallout4/Records/Major Records/PackageBranch.cs
```diff
@@ -15,6 +15,48 @@ public enum Flag
     }
 }
 
+public static class ProcedureType
```
Noggog, line 18: What are these classes used for?  Analyzers downstream?

Elscrux: Yeah these are just the types of strings that are "allowed" in the game. Sort of like an enum, but it is a string so this was my best thought how to implement this. Lmk if you know a better way though!

Mutagen.Bethesda.Fallout4/Records/Major Records/Package.xml
Elscrux: That's true, though nothing but a Keyword really makes sense for a linked reference. Either or some keyword.
https://en.uesp.net/wiki/Skyrim_Mod:Mod_File_Format/PACK also speaks about a keyword specifically.
Checking all vanilla files also shows that only keyword references are used here.

Mutagen.Bethesda.Fallout4/Records/Major Records/PackageBranch.cs
Noggog: They look like they're proper C# member names... so could just do an Enum with those values as fields.  And then you could just ToString to get the string value, and conveniently.. also the reverse and TryParse back to an Enum for validation of knowing a string is a ProcedureType

Noggog (approved):

PR #642, opened 2025-09-19
Fo4 PackageBranch ProcedureType and BranchType swapped to enum

Noggog:
Potentially an upgrade.   Can just ToString the enum.  Or TryParse to go the other way

### DIFF
--- a/Mutagen.Bethesda.Fallout4/Records/Major Records/PackageBranch.cs
+++ b/Mutagen.Bethesda.Fallout4/Records/Major Records/PackageBranch.cs
@@ -15,46 +15,46 @@ partial class PackageBranch
     }
 }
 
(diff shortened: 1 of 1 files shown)


Elscrux:
The plan is to change the types for ProcedureType and BranchType for all games to expose a struct member like this where we can add a direct conversion to an enum field to make it visible to users in the api directly. The plan is not to change it to an enum directly to also allow invalid configurations of this string to be loaded by Mutagen. Parsing the enum would just fail for these.
```
// An alias struct, which is just a wrapped string, but has a distinct type now
public record struct ProcedureType(string str)
{
    // conversions, etc.
}
```

Loqui doesn't support this kind of wrapper type atm so this will have to wait until then.

PR #643, opened 2025-09-30
VoiceTypeLookup: Fix leveled npcs contributing to GetIsRace and GetIsSex conditions

Elscrux:
Leveled npcs that potentially have a certain race or sex contributing all their potential voice types of races they could be to the GetIsRace condition. This cannot be the case for leveled npcs as the voice type is tied to the race and sex.

### DIFF
--- a/Mutagen.Bethesda.Skyrim/Records/Assets/VoiceType/VoiceTypeAssetLookup.cs
+++ b/Mutagen.Bethesda.Skyrim/Records/Assets/VoiceType/VoiceTypeAssetLookup.cs
@@ -15,7 +15,7 @@ public class VoiceTypeAssetLookup : IAssetCacheComponent
     private readonly Dictionary<FormKey, HashSet<FormKey>> _factionNPCs = new();
     private readonly Dictionary<FormKey, HashSet<FormKey>> _classNPCs = new();
     private readonly Dictionary<FormKey, HashSet<FormKey>> _raceNPCs = new();
-    private readonly Dictionary<bool, HashSet<FormKey>> _genderNPCs = new();
+    private readonly Dictionary<MaleFemaleGender, HashSet<FormKey>> _genderNPCs = new();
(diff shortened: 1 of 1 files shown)


Mutagen.Bethesda.Skyrim/Records/Assets/VoiceType/VoiceTypeAssetLookup.cs
```diff
@@ -941,31 +941,10 @@ private HashSet<bool> GetGenders(INpcGetter npc)
...
         {
-            return new HashSet<bool> { (npc.Configuration.Flags & NpcConfiguration.Flag.Female) != 0 };
+            return [(npc.Configuration.Flags & NpcConfiguration.Flag.Female) != 0];
         }
 
-        return npc.Template.IsNull ? new HashSet<bool>() : GetGenders(npc.Template);
-
-    }
-
-    private HashSet<bool> GetGenders(IFormLinkGetter<INpcSpawnGetter> npcTemplate)
```
Noggog, line 951: Somewhat unrelated to this PR, but just now noticing this `HashSet<bool>`  I think this is fairly confusing.  Is `true` male?    An enum Male/Female would be much better, even if just internal to this component.

Alternatively, can `Dictionary<bool, HashSet<FormKey>>` just be `GenderedItem<HashSet<FormKey>?>`?   Other places could also expose `GenderedItem<bool>` instead of the `HashSet<bool>`.

Noggog (changes requested):

Noggog (approved):

Mutagen.Bethesda.Skyrim/Records/Assets/VoiceType/VoiceTypeAssetLookup.cs
Elscrux: Good point, thanks. I changed bool to MaleFemaleGender.

PR #645, opened 2025-10-11
[FormKeyJsonConverter] Always write the type, even for generic types

Elscrux:
The problem is when the given type is a `IFormLinkGetter` without any generics which is set to a form link with an interface type like `IFormLinkGetter<INpcGetter>`. This would make the converter write a form key without any type info because it sees that the given type is actually a generic type even though the member type is just `IFormLinkGetter` .
But when trying to deserialize the form link again to a `IFormLinkGetter` it doesn't recognize this as a generic type and will expect a type info at the end of the string it reads, resulting in an exception.

Always adding the type info fixes this behavior, and due to Mutagen's handling of interface types this still works well. I hope this doesn't break anything else though! Tests seems to be fine after modifying them to the new behavior.

One change that there is due to this is that null form keys will also write a null form key with type, and not the actual null object.

### DIFF
--- a/Mutagen.Bethesda.Core.UnitTests/Json/JsonConverterTests.cs
+++ b/Mutagen.Bethesda.Core.UnitTests/Json/JsonConverterTests.cs
@@ -1,3 +1,4 @@
+using Loqui;
 using Shouldly;
 using Mutagen.Bethesda.Json;
 using Mutagen.Bethesda.Plugins;
@@ -16,6 +17,12 @@ public class JsonConverterTests
     class FormKeyClass
     {
         public FormKey Member { get; set; } = TestConstants.Form1;
+    }
+
+    static JsonConverterTests()
+    {
+        Warmup.Init();
+        LoquiRegistration.Register(TestMajorRecord_Registration.Instance);
     }
 
     [Fact]
@@ -174,7 +181,7 @@ public class JsonConverterTests
             Getter = new FormLink<ITestMajorRecordGetter>(TestConstants.Form2)
         };
         JsonConvert.SerializeObject(toSerialize, settings)
-            .ShouldBe($"{{\"Direct\":\"{toSerialize.Direct.FormKey}\",\"Setter\":\"{toSerialize.Direct.FormKey}\",\"Getter\":\"{toSerialize.Direct.FormKey}\"}}");
+            .ShouldBe($"{{\"Direct\":\"{toSerialize.Direct.FormKey}<TestGame.TestMajorRecord>\",\"Setter\":\"{toSerialize.Direct.FormKey}<TestGame.TestMajorRecord>\",\"Getter\":\"{toSerialize.Direct.FormKey}<TestGame.TestMajorRecord>\"}}");
     }
 
     [Fact]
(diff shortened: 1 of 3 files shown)


Noggog:
Sorry for the long delay.  Took me a bit to understand what was going on here when only glancing

Does this PR look okay to you? https://github.com/Mutagen-Modding/Mutagen/pull/647
A bit less invasive but hopefully same goal

Elscrux:
> Sorry for the long delay. Took me a bit to understand what was going on here when only glancing
> 
> Does this PR look okay to you? #647 A bit less invasive but hopefully same goal

@Noggog I don't think this fixes the issue. This is still not introducing the record type in the generic arguments to the written string from what I can see, and just the form key itself. The problem is that the field that is serialized doesn't have a generic parameter, but the object assigned to the field can be a generic type which the converter serializes without type information because it things that can just be inferred when things are deserialized again. But seeing as the type we look at when deserializing is just the getter type without generic parameters and then it throws because it doesn't know what form link this should be.
So I think we'd always need to include the type info, or maybe have a system that intelligently checks what needs to be saved, based on the actual field type and not the type of the object that is assigned to the field.

Noggog:
Hmm, i thought the actual goal was to get the test working `FormLinkInformationConverter_FormLink_Serialize_Deserialize_Generic_Interface` that had the `IFormLinkGetter Interface` set to a generic `FormLink<ITestMajorRecordGetter>` where the reseralization would fail.

True, the changes overall are less than the original PR, but the test passes, and it does export the generic information in that situation as needed.  
image

This fixes the exception with less changes.   Are there more situations that require adjustment?  

Only other major thing I see is writing the generic information with the null exports?   I guess you're just trying to add the information to everything as sort of a mass "might as well" push?   I guess i could see how it could be useful.  There is some type erasure potential when exporting nulls

Elscrux:
Oh, yeah you're right it does actually write the type information when calling `ToString` on the form link. Is that behavior consistent for everything? I used `IFormLinkIdentifier.GetString` like the other switch case uses to get the string and as I didn't see that I thought that would not work.
But in that case I assume it's fine. I also saw tests failing in the new PR but that seems to be other the other changes that result in this, I think you'll probably need to take some more of the changes in my branch or so if you want to update that.

So all in all I guess it's the same thing, just that my branch also includes type info with null as well. Maybe there are also situations where that might be useful, so I'm actually thinking this would be good to just merge as it is?

Noggog:
Stewing on it a bit, i do wonder if we just export the generic all the time.  Some code would "lose" the type info during a serialization pass when a value happens to be null.   Just cuz the value is null doesnt mean the type info should be lost?

I dont have any actual usage that would care yet, but who knows what others might expect type consistency from the system.

Elscrux:
> Stewing on it a bit, i do wonder if we just export the generic all the time. Some code would "lose" the type info during a serialization pass when a value happens to be null. Just cuz the value is null doesnt mean the type info should be lost?
> 
> I dont have any actual usage that would care yet, but who knows what others might expect type consistency from the system.

I think that would make sense

PR #652, opened 2026-02-17
Added Enderal Special Edition GOG game type

hakasapl:
This is untested in my application using mutagen but should fill in all the paths mutagen needs


Noggog (approved):

Noggog:
Looks like GameRelease.cs is missing an edit.  Ill spruce it up on my side.  Thanks for the PR!

PR #661, opened 2026-04-24
Fix AssetLinkBuilder generated paths

kieranknowles1:
Required for https://github.com/Mutagen-Modding/Mutagen.Bethesda.Analyzers/pull/567 tests to pass

Use BaseFolder relative path in generated asset links to closer align with real data. Create files in the test's data directory rather than in C:\Data\

### DIFF
--- a/Mutagen.Bethesda.Testing/AutoData/AssetLinkBuilder.cs
+++ b/Mutagen.Bethesda.Testing/AutoData/AssetLinkBuilder.cs
@@ -1,8 +1,9 @@
-﻿using System;
+using System;
 using System.IO.Abstractions;
 using System.Reflection;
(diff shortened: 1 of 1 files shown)


Noggog (approved):

Noggog:
@kieranknowles1 thanks!  Definitely better way to do it

Noggog:
Hmm, test failures.   I think one other improvement is that within the AutoFixture realm, the IDataDirectoryProvider should provide a path with a `Data` folder.  Right now it returns one with `DataDirectory`.   Didn't matter before, probably.  But AssetLinks -really- want `Data` to be part of the path.  So probably worth swapping.

Ill make a PR to this PR

PR #664, opened 2026-05-16
Add ModelBehavior and ModelBodyTexture variants

Elscrux:


Elscrux:
I've separated the two model's with special asset types to their own classes.

Though it'd be good to check if everything still works as intended, for example this removes the IModeled interface from BodyData because it doesn't match the asset type anymore. That would be a breaking change.

Mutagen.Bethesda.Skyrim/Records/Common Subrecords/ModelBehavior.xml
```diff
@@ -0,0 +1,14 @@
+<?xml version="1.0" encoding="us-ascii"?>
+<Loqui xmlns="http://tempuri.org/LoquiSource.xsd">
+  <Object name="SimpleModelBehavior" objType="Subrecord">
```
Noggog: I think the distinction between SimpleModelBehavor and ModelBehavior isn't needed?  For Model it's needed as there's some variants that only have some parts.    But for Behaviors (and BodyTextures), it's pretty straightforward usage?   I think we can just merge together to ModelBehavior and ModelBodyTexture, no simple variants

Noggog (changes requested):

Mutagen.Bethesda.Skyrim/Records/Common Subrecords/ModelBehavior.xml
Elscrux: Sounds good

Noggog (approved):

PR #669, opened 2026-06-19
Load Order Merged Location Fields Extension Methods

Elscrux:
Add variants of location extensions which look at the load order to merge the states of special fields in the location record which are also merged at runtime

### DIFF
--- a/Mutagen.Bethesda.Core/Plugins/Cache/LinkCacheOverridesMixIn.cs
+++ b/Mutagen.Bethesda.Core/Plugins/Cache/LinkCacheOverridesMixIn.cs
@@ -4,6 +4,138 @@ namespace Mutagen.Bethesda.Plugins.Cache;
 
 public static class LinkCacheOverridesMixIn
 {
+	#region GetPreviousOverrideContexts
+	/// <summary>
+	/// Gets all previous overrides of a record in the link cache, based on the specified override mod key and resolve target.
+	/// </summary>
+	/// <param name="cache">Link cache to search for previous overrides</param>
+	/// <param name="record">The record for which to find previous overrides</param>
+	/// <param name="overrideModKey">The mod key of the override to start searching from</param>
+	/// <param name="target">Resolution target to look up previous overrides from</param>
+	/// <typeparam name="TMod">The type of Mod to look up</typeparam>
+	/// <typeparam name="TModGetter">The getter type of Mod to look up</typeparam>
+	/// <typeparam name="TMajor">The type of Major Record to look up</typeparam>
+	/// <typeparam name="TMajorGetter">The getter type of Major Record to look up</typeparam>
+	/// <returns>Enumerable of all previous overrides of the record, based on the specified override mod key and resolve target</returns>
+	public static IEnumerable<IModContext<TMod, TModGetter, TMajor, TMajorGetter>> GetPreviousOverrideContexts<TMod, TModGetter, TMajor, TMajorGetter>(this ILinkCache<TMod, TModGetter> cache, TMajor record, ModKey overrideModKey, ResolveTarget target = ResolveTarget.Winner)
+		where TMod : class, TModGetter, IMod
+		where TModGetter : class, IModGetter
+		where TMajor : class, TMajorGetter, IMajorRecord
+		where TMajorGetter : class, IMajorRecordGetter
+	{
+		return target switch
+		{
+			ResolveTarget.Winner => cache.ResolveAllContexts<TMajor, TMajorGetter>(record)
+				.SkipWhile(x => x.ModKey != overrideModKey)
+				.Skip(1),
+			_ => cache.ResolveAllContexts<TMajor, TMajorGetter>(record, ResolveTarget.Origin)
+				.TakeWhile(x => x.ModKey != overrideModKey)
+		};
+	}
+
+	/// <summary>
+	/// Gets all previous overrides of a record in the link cache, based on the specified override mod key and resolve target.
+	/// </summary>
+	/// <param name="cache">Link cache to search for previous overrides</param>
+	/// <param name="formKey">FormKey to find previous overrides for</param>
+	/// <param name="overrideModKey">The mod key of the override to start searching from</param>
+	/// <param name="target">Resolution target to look up previous overrides from</param>
+	/// <typeparam name="TMod">The type of Mod to look up</typeparam>
+	/// <typeparam name="TModGetter">The getter type of Mod to look up</typeparam>
+	/// <typeparam name="TMajor">The type of Major Record to look up</typeparam>
+	/// <typeparam name="TMajorGetter">The getter type of Major Record to look up</typeparam>
+	/// <returns>Enumerable of all previous overrides of the record, based on the specified override mod key and resolve target</returns>
+	public static IEnumerable<IModContext<TMod, TModGetter, IMajorRecord, IMajorRecordGetter>> GetPreviousOverrideContexts<TMod, TModGetter, TMajor, TMajorGetter>(this ILinkCache<TMod, TModGetter> cache, FormKey formKey, ModKey overrideModKey, ResolveTarget target = ResolveTarget.Winner)
+		where TMod : class, TModGetter, IMod
+		where TModGetter : class, IModGetter
+		where TMajor : class, TMajorGetter, IMajorRecord
+		where TMajorGetter : class, IMajorRecordGetter
+	{
+		return target switch
+		{
+			ResolveTarget.Winner => cache.ResolveAllContexts<TMajor, TMajorGetter>(formKey)
+				.SkipWhile(x => x.ModKey != overrideModKey)
+				.Skip(1),
+			_ => cache.ResolveAllContexts<TMajor, TMajorGetter>(formKey, ResolveTarget.Origin)
+				.TakeWhile(x => x.ModKey != overrideModKey)
+		};
+	}
+
+	/// <summary>
+	/// Gets all previous overrides of a record in the link cache, based on the specified override mod key and resolve target.
+	/// </summary>
+	/// <param name="cache">Link cache to search for previous overrides</param>
+	/// <param name="formLink">FormLink to find previous overrides for</param>
+	/// <param name="overrideModKey">The mod key of the override to start searching from</param>
+	/// <param name="target">Resolution target to look up previous overrides from</param>
+	/// <typeparam name="TMod">The type of Mod to look up</typeparam>
+	/// <typeparam name="TModGetter">The getter type of Mod to look up</typeparam>
+	/// <returns>Enumerable of all previous overrides of the record, based on the specified override mod key and resolve target</returns>
+	public static IEnumerable<IModContext<TMod, TModGetter, IMajorRecord, IMajorRecordGetter>> GetPreviousOverrideContexts<TMod, TModGetter>(this ILinkCache<TMod, TModGetter> cache, IFormLinkIdentifier formLink, ModKey overrideModKey, ResolveTarget target = ResolveTarget.Winner)
+		where TMod : class, TModGetter, IMod
+		where TModGetter : class, IModGetter
+	{
+		return target switch
+		{
+			ResolveTarget.Winner => cache.ResolveAllContexts(formLink)
+				.SkipWhile(x => x.ModKey != overrideModKey)
+				.Skip(1),
+			_ => cache.ResolveAllContexts(formLink, ResolveTarget.Origin)
+				.TakeWhile(x => x.ModKey != overrideModKey)
+		};
+	}
+
+	/// <summary>
+	/// Gets all previous overrides of a record in the link cache, based on the specified override mod key and resolve target.
+	/// </summary>
+	/// <param name="cache">Link cache to search for previous overrides</param>
+	/// <param name="formKey">FormKey to find previous overrides for</param>
+	/// <param name="type">The type of record to look up</param>
+	/// <param name="overrideModKey">The mod key of the override to start searching from</param>
+	/// <param name="target">Resolution target to look up previous overrides from</param>
+	/// <typeparam name="TMod">The type of Mod to look up</typeparam>
+	/// <typeparam name="TModGetter">The getter type of Mod to look up</typeparam>
+	/// <returns>Enumerable of all previous overrides of the record, based on the specified override mod key and resolve target</returns>
+	public static IEnumerable<IModContext<TMod, TModGetter, IMajorRecord, IMajorRecordGetter>> GetPreviousOverrideContexts<TMod, TModGetter>(this ILinkCache<TMod, TModGetter> cache, FormKey formKey, Type type, ModKey overrideModKey, ResolveTarget target = ResolveTarget.Winner)
+		where TMod : class, TModGetter, IMod
+		where TModGetter : class, IModGetter
+	{
+		return target switch
+		{
+			ResolveTarget.Winner => cache.ResolveAllContexts(formKey, type)
+				.SkipWhile(x => x.ModKey != overrideModKey)
+				.Skip(1),
+			_ => cache.ResolveAllContexts(formKey, type, ResolveTarget.Origin)
+				.TakeWhile(x => x.ModKey != overrideModKey)
+		};
+	}
+
+	/// <summary>
+	/// Gets all previous overrides of a record in the link cache, based on the specified override mod key and resolve target.
+	/// </summary>
+	/// <param name="cache">Link cache to search for previous overrides</param>
+	/// <param name="formKey">FormKey to find previous overrides for</param>
+	/// <param name="overrideModKey">The mod key of the override to start searching from</param>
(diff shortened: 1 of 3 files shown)


Mutagen.Bethesda.Skyrim/Extensions/LocationExt.cs
```diff
@@ -33,6 +35,41 @@ IEnumerable<ILocationRefTypeReferenceGetter> Added()
             .Where(x => !removed.Contains(x.Ref.FormKey));
     }
 
+    public static IEnumerable<ILocationRefTypeReferenceGetter> AllLocationRefTypesReferences(this ILocationGetter location, ILinkCache linkCache)
+    {
+        var previousOverrides = linkCache
+            .ResolveAll(location, ResolveTarget.Origin)
+            .TakeWhile(x => !Equals(x, location))
```
Noggog: Hmm, this equality check is checking every field one by one.  Is this intentional?   Also, this equals feature isnt complete or tested accurate.

Mutagen.Bethesda.Skyrim/Extensions/LocationExt.cs
```diff
@@ -33,6 +35,41 @@ IEnumerable<ILocationRefTypeReferenceGetter> Added()
...
+            .Append(location)
+            .ToArray();
+
+        var refTypes = new HashSet<ILocationRefTypeReferenceGetter>();
+        foreach (var previousOverride in previousOverrides)
+        {
+            refTypes.Add(Added(previousOverride));
+            var removed = Removed(previousOverride);
+            refTypes.RemoveWhere(x => removed.Any(r => r.FormKey == x.Ref.FormKey));
```
Noggog: Maybe put the removed formkeys in a hash set and do a contains here

Mutagen.Bethesda.Skyrim/Extensions/LocationExt.cs
```diff
@@ -63,6 +100,41 @@ IEnumerable<IPersistentActorReferenceGetter> Added()
             .Where(x => !removed.Contains(x.Actor.FormKey));
     }
 
+    public static HashSet<IPersistentActorReferenceGetter> AllPersistentActorReferences(this ILocationGetter location, ILinkCache linkCache)
```
Noggog: Are we really returning a HashSet of the actual objects wholesale?   Seems wrong

Mutagen.Bethesda.Skyrim/Extensions/LocationExt.cs
Elscrux: Yeah it's probably not the best. What I want is to get all previous override of a record. If Mutagen could also support that feature natively that would be huge of course. Currently it's guessed based on having the exact same content though I see that that's not ideal. I assume that getting all previous overrides would require us to know the ModKey of the current override right? So we need to resolve the context of ours somehow? I'm honestly not sure what is the best thing to do here. Mutagen seems fairly limited when it comes to specific override stuff, also looking at that other PR for analyzers https://github.com/Mutagen-Modding/Mutagen.Bethesda.Analyzers/pull/608.

Mutagen.Bethesda.Skyrim/Extensions/LocationExt.cs
Elscrux: Good idea!

Mutagen.Bethesda.Skyrim/Extensions/LocationExt.cs
Elscrux: Oh, yeah that's oversight, thanks!

Mutagen.Bethesda.Skyrim/Extensions/LocationExt.cs
Elscrux: Thinking about stuff like this rn.
But not sure if there aren't maybe better and more performant ways to do it? Something that is handled internally in the link cache?
```cs
public IModContext<T> GetSpecificOverride<T>(T record, ModKey overrideModKey) where T : class, IMajorRecordGetter
{
	return linkCache.ResolveAllSimpleContexts(record)
			.FirstOrDefault(x => x.ModKey == overrideModKey)
	 ?? throw new KeyNotFoundException($"No override of {record.FormKey} found in {overrideModKey}");
}
public IEnumerable<IModContext<T>> GetPreviousOverrides<T>(T record, ModKey overrideModKey) where T : class, IMajorRecordGetter
{
	return linkCache.ResolveAllSimpleContexts(record)
		.TakeWhile(x => x.ModKey != overrideModKey);
}
```

Noggog: Hmm, by definition, all overrides of a record would share the same FormKey.  So just looking through all the mods for the record with the same FormKey would find all the overrides.

```cs
        var previousOverrides = linkCache
            .ResolveAll(location, ResolveTarget.Origin)
```

This already is finding all the overrides in the load order?   (you have it starting from the origin moving forward, but it'll iterate them all eventually)  There's no need to filter it down, i dont think?   At least, it supposed to for more normal records.  Maybe it's having issues finding them for oddball records like Landscape.

But for, say, an NPC with formkey `123456:MyMod.esp`, it'll find all copies of npc records with Formkey `123456:MyMod.esp` from all mods in the LO

Elscrux: I think we have a misunderstanding, I probably didn't explain it properly.

Imagine a load order like this:
0: Skyrim.esm
1: Update.esm
2. MyMod.esm
3. SomeOtherMod.esm

So what I currently set it up to be is that when you call `location.AllLocationRefTypesReferences(linkCache)` where the location is from `MyMod.esm`, then it would give you the location fields from Skyrim.esm, Update.esm + MyMod.esm and merge them all together like it happens in game to then return the state of the fields that they would be in the final game _up to that point with MyMod.esm_. I didn't set it up to be evaluating the whole load order but to be evaluating it for a specific record override that you pass in. That is because I also want to use this function in analyzers where I might pass in the record of some mod, not necessarily the winning override. If you would always want the winning override (and thus all previous overrides, so you could use `ResolveAll()`). But in this case I actually just want the previous overrides from MyMod.esm.

Noggog: I think i follow.    You have Locations X, Y, Z + tons of others.  And you want to say, get all overrides of X as they appear from Skyrim.esm, Update.esm.   But exclude SomeOtherMod.esm, because that's after MyMod.esm, and thus not a previous override.

I was confused thinking the Equals check was also intended to confirm it was the same location record X and not Y or Z

Noggog:
GetSpecificOverride has been added recently in that other PR as TryResolveForMod?

--------

Here's maybe how i'd write GetPreviousOverrides

```cs
public IEnumerable<IModContext<T>> GetPreviousOverrides<T>(T record, ModKey overrideModKey) where T : class, IMajorRecordGetter
{
	return linkCache.ResolveAllSimpleContexts(record, Target.Winning)
		.SkipWhile(x => x.ModKey != overrideModKey) // Skip until we find target mod
                .Skip(1); // Skip target mod itself
                // Return the rest
}
```

This has the benefit of iterating "winningmost oriented", which users will expect, generally.   Eg, they'd get Update.esm first, then Skyrim.esm finally as the origin last.

Whether we need to add this officially to the load order.. hmm.   I dont think it would get much more optimized than this.   Any caching would just be caching a few modkey checks to skip, not too consequential

I think that's probably fine enough to add as an extension function on ILinkCache rather than actually adding it to the interface?

Elscrux:
> GetSpecificOverride has been added recently in that other PR as TryResolveForMod?
> 
> Here's maybe how i'd write GetPreviousOverrides
> 
> ```cs
> public IEnumerable<IModContext<T>> GetPreviousOverrides<T>(T record, ModKey overrideModKey) where T : class, IMajorRecordGetter
> {
> 	return linkCache.ResolveAllSimpleContexts(record, Target.Winning)
> 		.SkipWhile(x => x.ModKey != overrideModKey) // Skip until we find target mod
>                 .Skip(1); // Skip target mod itself
>                 // Return the rest
> }
> ```
> 
> This has the benefit of iterating "winningmost oriented", which users will expect, generally. Eg, they'd get Update.esm first, then Skyrim.esm finally as the origin last.
> 
> Whether we need to add this officially to the load order.. hmm. I dont think it would get much more optimized than this. Any caching would just be caching a few modkey checks to skip, not too consequential
> 
> I think that's probably fine enough to add as an extension function on ILinkCache rather than actually adding it to the interface?

Sorry I forgot to comment here!

Putting it as an extension function on ILinkCache is definitely good. And as for the implementation, do we want to expose both orders? And use the default order to get it in the winning order?

Elscrux:
@Noggog We still have an unresolved problem here I realize.
Looking at the calls we want to replace:
```cs
        var previousOverrides = linkCache
            .ResolveAll(location, ResolveTarget.Origin)
            .TakeWhile(x => !Equals(x, location))
            .Append(location)
            .ToArray();
```

Even with our GetPreviousOverrides we're not able to say from what point we want the previous overrides for. This is a bigger challenge in general seeing as even the current way with equality isn't really fool proof as multiple overrides can have the exact same members set and thus matching equality.

So I assume we always need to pass the ModKey of the override we want into these functions? Or is there some way in Mutagen how we could track what mod a specific record is from? Not currently I'm pretty sure but would that otherwise be anything that would have a use?
```cs
public static IEnumerable<ILocationRefTypeReferenceGetter> AllLocationRefTypesReferences(
        this ILocationGetter location,
        ILinkCache linkCache)
// Add ModKey as parameter
```

Mutagen.Bethesda.Skyrim/Extensions/LocationExt.cs
```diff
@@ -33,6 +35,41 @@ IEnumerable<ILocationRefTypeReferenceGetter> Added()
             .Where(x => !removed.Contains(x.Ref.FormKey));
     }
 
+    public static IEnumerable<ILocationRefTypeReferenceGetter> AllLocationRefTypesReferences(this ILocationGetter location, ILinkCache linkCache)
+    {
+        var previousOverrides = linkCache
```
Noggog: Been a bit so im fuzzy, but I think we can swap this to use the GetPreviousOverrides call?

Noggog (commented):
> Even with our GetPreviousOverrides we're not able to say from what point we want the previous overrides for

What do you mean exactly?  We're able to pass the modkey we want to start getting overrides from.

```
ModA
...
ModZ
```

If we pass GetPreviousOverrides(ModF), we'd get overrides only from mods A-E.  Is that not the ability to control what point we want overrides for?

Or are you saying that extension method isn't "true", as it's not embedded directly in the link cache itself?  I think the extension method is fine enough, myself.  We can add to the interface explicitly if we think the optimization is worth it.  I think the `TakeWhile(!=ModKey)` is efficient enough for the complexity tradeoffs.

---------

Also, i think I'd really like to see more unit tests for all these.   Skipping over them on past utility functions is already causing problems for kieran over in #679 where intentions of what to include exactly are unclear, and the optimization work might very well have regressions and different return values now that we can't easily detect.

Elscrux:
@Noggog sorry, I probably didn't make it very clear. My point was that just passing in a record means you have no way of knowing what mod that record belonged to. Just doing `Equals` to check for a record won't do it because one record might have the same equality for multiple mods. So we can't do `GetPreviousOverrides(IMajorRecordGetter record)`, we have to do `GetPreviousOverrides(ModKey, IFormLinkGetter formLink)`. Or is there a better way?

Noggog:
I think passing in the ModKey is the best way, yes.  A record object alone has no idea which mod contained it (in fact, an object in theory could have been added to several mod objects).   The record is the payload value.  But it doesn't have the information to be its own key in this situation.

Elscrux:
Great, I think we should be pretty much ready now! Lmk if there's anything else you spot :)

Mutagen.Bethesda.Core/Plugins/Cache/LinkCacheOverridesMixIn.cs
```diff
@@ -98,7 +99,7 @@ public static IEnumerable<IModContext> GetPreviousOverrides(this ILinkCache cach
 	/// <param name="target">Resolution target to look up previous overrides from</param>
 	/// <returns>Enumerable of all previous overrides of the record, based on the specified override mod key and resolve target</returns>
 	[Obsolete("This call is not as optimized as its generic typed counterpart.  Use as a last resort.")]
-	public static IEnumerable<IModContext> GetPreviousOverrides(this ILinkCache cache, FormKey formKey, ModKey overrideModKey, ResolveTarget target = ResolveTarget.Winner)
+	public static IEnumerable<IModContext<IMajorRecordGetter>> GetPreviousOverridesContext(this ILinkCache cache, FormKey formKey, ModKey overrideModKey, ResolveTarget target = ResolveTarget.Winner)
```
Noggog: I think `GetPreviousOverrideContexts` would be better?    Plural being that we're looping contexts (that are overrides)

Mutagen.Bethesda.Core/Plugins/Cache/LinkCacheOverridesMixIn.cs
```diff
@@ -110,4 +111,82 @@ public static IEnumerable<IModContext> GetPreviousOverrides(this ILinkCache cach
...
+	/// <param name="cache">Link cache to search for previous overrides</param>
+	/// <param name="record">The record for which to find previous overrides</param>
+	/// <param name="overrideModKey">The mod key of the override to start searching from</param>
+	/// <param name="target">Resolution target to look up previous overrides from</param>
+	/// <typeparam name="TMajor">The type of Major Record to look up</typeparam>
+	/// <returns>Enumerable of all previous overrides of the record, based on the specified override mod key and resolve target</returns>
+	public static IEnumerable<TMajor> GetPreviousOverrides<TMajor>(this ILinkCache cache, TMajor record, ModKey overrideModKey, ResolveTarget target = ResolveTarget.Winner)
+		where TMajor : class, IMajorRecordGetter
+	{
+		return cache.GetPreviousOverridesContext(record, overrideModKey, target)
```
Noggog: The GetPreviousOverrides using GetPreviousOverridesContext is fine, as it uses simple contexts (eg, doesn't spin up all the wiring needed to GetOrAddAsOverride, which can be complex)

But I do wonder if we should add GetPreviousOverrideContexts alternatives that DO give complex contexts back (eg, the full 4 generics alternative).  Probably good to do for completeness.

Noggog (changes requested):

Mutagen.Bethesda.Core/Plugins/Cache/LinkCacheOverridesMixIn.cs
Elscrux: Good point!

Mutagen.Bethesda.Core/Plugins/Cache/LinkCacheOverridesMixIn.cs
Elscrux: Do you have a preference? I assume in that case we'd have GetPreviousOverrides, GetPreviousOverridesSimpleContexts, and GetPreviousOverridesContexts.

Noggog: GetPreviousOverrides for records

GetPreviousOverrideContexts for contexts?  I'd prefer only one plural, and it feels more appropriate at the end.  "im getting the contexts that are overriding".     The above one could've been named GetPreviousOverrideRecords "im getting the records that are overriding", in the same way, but Record is implicit and omitted, so the plural has to go on Overrides?

`GetPreviousSimpleOverrideContexts` for the simple contexts

Elscrux: Yeah for sure. I'm just quite sure about `GetPreviousSimpleOverrideContexts`  because it doesn't sort next to the other `GetPreviousOverride` methods, making it a little less easy to see.
But it's good to know you want the three different variants, I can add those.

Elscrux: Actually there doesn't seem to be a ResolveAllContexts, just ResolveAllSimpleContexts. I assume we'd want that before anything?

Noggog: ResolveAllContexts should exist
https://github.com/Mutagen-Modding/Mutagen/blob/dev/Mutagen.Bethesda.Core/Plugins/Cache/ILinkCache.cs#L1454

Maybe something in the way you're calling it isn't lining up?

Elscrux: Ah yes, the link cache needed generic parameters. Simply `ILinkCache` wasn't enough.

I've adjusted everything now, does that work for you?

Noggog (approved):

PR #670, opened 2026-06-20
ILinkCache.TryGetLinkCacheForMod

Noggog:
Adds the ability to get an ILinkCache that is scoped to a single mod from an existing link cache.

First user: https://github.com/Mutagen-Modding/Mutagen.Bethesda.Analyzers/pull/608

### DIFF
--- a/Mutagen.Bethesda.Core/Mutagen.Bethesda.Core.csproj
+++ b/Mutagen.Bethesda.Core/Mutagen.Bethesda.Core.csproj
@@ -653,6 +653,7 @@
         <Compile Include="Plugins\FormLinkNullable.cs" />
         <Compile Include="Plugins\IFormLink.cs" />
         <Compile Include="Plugins\Cache\ILinkCache.cs" />
+        <Compile Include="Plugins\Cache\LinkCacheFromModMixIn.cs" />
         <Compile Include="Plugins\Cache\LinkCachePreferences.cs" />
         <Compile Include="Plugins\FormLinkTypelessComparer.cs" />
         <Compile Include="Plugins\Order\LoadOrder.cs" />
--- a/Mutagen.Bethesda.Core/Plugins/Cache/ILinkCache.cs
+++ b/Mutagen.Bethesda.Core/Plugins/Cache/ILinkCache.cs
@@ -131,8 +131,7 @@ public interface ILinkCache : IIdentifierLinkCache, IWinningOverrideProvider
     /// </exception>
     /// <returns>True if a matching record was found</returns>
     bool TryResolve(FormKey formKey, Type type, [MaybeNullWhen(false)] out IMajorRecordGetter majorRec, ResolveTarget target = ResolveTarget.Winner);
-
-
+    
     /// <summary>
     /// Retrieves the winning record that matches the FormKey relative to the source the package was attached to.<br/>
     /// <br/>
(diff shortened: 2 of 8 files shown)


Elscrux (commented):
Thanks a lot! Looking pretty good :) And wow that's a lot of boilerplate code needed to add this, whoops sorry :D
One thing I think that might be missing is the ResolveIdentifier function which we will need for https://github.com/Mutagen-Modding/Mutagen.Bethesda.Analyzers/pull/608 right?

Also please let me know if you think we should also add a a call like `ResolveAllPreviousOverrides(FormLink formLink, ModKey modKey)? Also related to this https://github.com/Mutagen-Modding/Mutagen/pull/669#discussion_r3445670508 but it could also be useful in other places.

Noggog:
> ResolveAllPreviousOverrides

Copied from your link
```cs
public IEnumerable<IModContext<T>> GetPreviousOverrides<T>(T record, ModKey overrideModKey) where T : class, IMajorRecordGetter
{
	return linkCache.ResolveAllSimpleContexts(record)
		.TakeWhile(x => x.ModKey != overrideModKey);
}
```

I wouldn't call that previous overrides?  Seems more like an "All overrides except X"?  Which, i guess if X is the winningmost is coincidentally all previous overrides.

But I think just looping them all and skipping is performant enough.  The linkcache wont be doing anything fancier than that anyway.

Can chat about it more on another issue/PR if we really want it.  Gonna get this into the nightly

Elscrux:
> > ResolveAllPreviousOverrides
> 
> Copied from your link
> 
> ```cs
> public IEnumerable<IModContext<T>> GetPreviousOverrides<T>(T record, ModKey overrideModKey) where T : class, IMajorRecordGetter
> {
> 	return linkCache.ResolveAllSimpleContexts(record)
> 		.TakeWhile(x => x.ModKey != overrideModKey);
> }
> ```
> 
> I wouldn't call that previous overrides? Seems more like an "All overrides except X"? Which, i guess if X is the winningmost is coincidentally all previous overrides.

It's now `While()` but `TakeWhile()`. So it should be previous overrides unless I got something totally wrong.

> But I think just looping them all and skipping is performant enough. The linkcache wont be doing anything fancier than that anyway.
> 
> Can chat about it more on another issue/PR if we really want it. Gonna get this into the nightly

We can add it as extension function or so if wanted maybe? Then we could always modify how it works under the hood if improvements can be made to it after all.

Noggog:
Hmm, consider record A in mods W, X, Y, Z, with that being the load order, Z being the winningmost

You give that call record A, override ModKey Y

I think it would return just Z to you.  It loops simple contexts, starting with winning override (Z), returns it.  Then hits Y and then stops.

I think you'd want to mix in a target of Origin to the Resolve call, so it starts from W (the origin mod), then returns X, then stops at Y.

Even then, some people might expect the returning order to be X then W?   Return the previous overrides, yeah, but still in winningmost order.

Anyway, a chat for another card, perhaps

Elscrux:
> Hmm, consider record A in mods W, X, Y, Z, with that being the load order, Z being the winningmost
> 
> You give that call record A, override ModKey Y
> 
> I think it would return just Z to you. It loops simple contexts, starting with winning override (Z), returns it. Then hits Y and then stops.
> 
> I think you'd want to mix in a target of Origin to the Resolve call, so it starts from W (the origin mod), then returns X, then stops at Y.
> 
> Even then, some people might expect the returning order to be X then W? Return the previous overrides, yeah, but still in winningmost order.
> 
> Anyway, a chat for another card, perhaps

@Noggog Oh, yes sorry I meant to be saying that we want to take the overrides starting from the origin.

See here the code from #669 which does this. If you want I can add it there. It would basically just be a shortcut unless there would be a better way to implement it https://github.com/Elscrux/Mutagen/blob/629f847a1bede733d7e990ce0cd6b5af001bae46/Mutagen.Bethesda.Skyrim/Extensions/LocationExt.cs#L40-L42

PR #679, opened 2026-07-25
VoiceTypeAssetLookup optimisations

kieranknowles1:
Optimizations and refactoring of VoiceTypeAssetLookup

- Reduce usage of HashSet during prepare
- Build lookups from winning overrides
- Cache voice type -> speakers for cases where the whole voice type is used
- Use Lazy for child NPCs, scene aliases, and shared info usages
- Use enumerable for SpeakerVoices value as most NPCs only have one voice type.

Scene aliases and shared info usages may benefit further from a usage cache, this is not implemented here as it would be a breaking change.

Overall, this gives a 1.8x speedup and reduction in allocations when preparing the lookup against the vanilla masters and a
45x speedup with 5x reduction in allocations when processing GetIsVoiceType conditions.

---
Before:
| Method                      | Mean         | Error      | StdDev     | Gen0    | Gen1 | Gen2  | Allocated |
|---------------------------- | ------------:|-----------:|-----------:|--------:|-------:|-:|----------:|
| PrepareWithDefaultMasters | 261.9 ms | 5.21 ms |  4.87 ms | 17000.0000 | 5000.0000 | 1000.0000 | 282.92 MB |
| LookupVoicesSpecificSpeaker |     2.384 us |  0.0473 us |  0.0790 us |  0.4158 |      - | - |    6.8 KB |
| LookupVoicesGenericSpeaker  | 5,460.695 us | 49.8194 us | 41.6015 us | 15.6250 | 7.8125 | - | 379.58 KB |

After:
| Method                      | Mean       | Error     | StdDev    | Gen0   | Gen1 | Gen2  | Allocated |
|------------------------ |-----------:|----------:|----------:|-------:|-------:|-:|----------:|
| PrepareWithDefaultMasters | 144.8 ms | 2.53 ms | 3.19 ms | 9000.0000 | 2000.0000 |        - | 157.56 MB |
| LookupVoicesSpecificSpeaker |   2.178 us | 0.0197 us | 0.0175 us | 0.4005 |      - | - |   6.59 KB |
| LookupVoicesGenericSpeaker | 119.337 us | 0.6232 us | 0.5204 us | 4.0283 | 0.7324 | - |  67.18 KB |

### DIFF
--- a/Mutagen.Bethesda.Skyrim/Records/Assets/VoiceType/VoiceTypeAssetLookup.cs
+++ b/Mutagen.Bethesda.Skyrim/Records/Assets/VoiceType/VoiceTypeAssetLookup.cs
@@ -31,136 +31,126 @@ public class VoiceTypeAssetLookup : IAssetCacheComponent
     {
         _formLinkCache = linkCache.FormLinkCache;
 
-        var childRaces = new HashSet<FormKey>();
-        foreach (var mod in _formLinkCache.PriorityOrder)
+        foreach (var quest in _formLinkCache.WinningOverrides<IQuestGetter>())
         {
-            foreach (var quest in mod.EnumerateMajorRecords<IQuestGetter>())
+            foreach (var alias in quest.Aliases)
             {
-                foreach (var alias in quest.Aliases)
-                {
-                    var uniqueActor = alias.UniqueActor.FormKey;
-                    if (uniqueActor.IsNull) continue;
+                var uniqueActor = alias.UniqueActor.FormKey;
+                if (uniqueActor.IsNull) continue;
 
-                    foreach (var faction in alias.Factions)
+                foreach (var faction in alias.Factions)
+                {
+                    if (!faction.IsNull)
                     {
-                        if (!faction.IsNull)
-                        {
-                            _factionNPCs
-                                .GetOrAdd(faction.FormKey)
-                                .Add(uniqueActor);
-                        }
-                    }
-                }
-            }
-
-            foreach (var leveledNpc in mod.EnumerateMajorRecords<ILeveledNpcGetter>())
-            {
-                if (leveledNpc.Entries is null) continue;
-
-                var voiceTypes = leveledNpc.Entries
-                    .Select(x => x.Data?.Reference)
-                    .WhereNotNull()
-                    .SelectMany(GetVoiceTypes)
-                    .ToHashSet();
-
-                _speakerVoices
-                    .GetOrAdd(leveledNpc.FormKey)
-                    .Add(voiceTypes);
-            }
-
-            foreach (var npc in mod.EnumerateMajorRecords<INpcGetter>())
-            {
-                _speakerVoices.GetOrAdd(npc.FormKey, () => GetVoiceTypes(npc).ToHashSet());
-
-                foreach (var factionKey in GetFactions(npc))
-                {
-                    _factionNPCs
-                        .GetOrAdd(factionKey.FormKey)
-                        .Add(npc.FormKey);
-                }
-
-                foreach (var classKey in GetClasses(npc))
-                {
-                    _classNPCs
-                        .GetOrAdd(classKey.FormKey)
-                        .Add(npc.FormKey);
-                }
-
-                foreach (var gender in GetGenders(npc))
-                {
-                    _genderNPCs
-                        .GetOrAdd(gender)
-                        .Add(npc.FormKey);
-                }
-
-                foreach (var raceKey in GetRaces(npc))
-                {
-                    _raceNPCs
-                        .GetOrAdd(raceKey.FormKey)
-                        .Add(npc.FormKey);
-                }
-            }
-
-            foreach (var response in mod.EnumerateMajorRecords<IDialogResponsesGetter>())
-            {
-                if (!response.ResponseData.IsNull)
-                {
-                    _sharedInfoUsages
-                        .GetOrAdd(response.ResponseData.FormKey)
-                        .Add(response.FormKey);
-                }
-            }
-
-            foreach (var talkingActivator in mod.EnumerateMajorRecords<ITalkingActivatorGetter>())
-            {
-                if (!_speakerVoices.ContainsKey(talkingActivator.FormKey))
-                {
-                    var voice = GetVoiceType(talkingActivator);
-                    if (voice != null)
-                        _speakerVoices.Add(talkingActivator.FormKey, [voice]);
-                }
-            }
-
-            foreach (var race in mod.EnumerateMajorRecords<IRaceGetter>())
-            {
-                if ((race.Flags & Race.Flag.Child) != 0)
-                {
-                    childRaces.Add(race.FormKey);
-                }
-            }
-
-            foreach (var scene in mod.EnumerateMajorRecords<ISceneGetter>())
-            {
-                foreach (var action in scene.Actions)
-                {
-                    if (action.Type == SceneAction.TypeEnum.Dialog && !action.Topic.IsNull && action.ActorID != null && !_dialogueSceneAliasIndex.ContainsKey(action.Topic.FormKey))
-                    {
-                        _dialogueSceneAliasIndex.Add(action.Topic.FormKey, action.ActorID.Value);
-                    }
-                }
-            }
-
-            var defaultVoiceTypes = new HashSet<string>();
-            _defaultVoiceTypes.Add(mod.ModKey, defaultVoiceTypes);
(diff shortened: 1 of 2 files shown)


Mutagen.Bethesda.Skyrim/Records/Assets/VoiceType/VoiceTypeAssetLookup.cs
```diff
@@ -12,14 +12,18 @@ public class VoiceTypeAssetLookup : IAssetCacheComponent
 
     //Databases
     private readonly Dictionary<ModKey, HashSet<string>> _defaultVoiceTypes = new();
-    private readonly Dictionary<FormKey, HashSet<string>> _speakerVoices = new();
+    // NPC -> Voice types
+    private readonly Dictionary<FormKey, IEnumerable<string>> _speakerVoices = new();
```
Noggog, line 16: Why make the value type here IEnumerable<string> when we're filling it with a HashSet<string> on line 67?

Doesn't matter too much, since it's all private.. but seems like an unnecessary change

Mutagen.Bethesda.Tests/Benchmarks/VoiceTypeAssetLookupBenchmark.cs
```diff
@@ -0,0 +1,68 @@
...
+    {
+        var env = GameEnvironment.Typical.Builder<ISkyrimMod, ISkyrimModGetter>(GameRelease.SkyrimSE)
+            .Build();
+        LinkCache = env.LinkCache;
+        Lookup = new();
+        var assetCache = LinkCache.CreateImmutableAssetLinkCache();
+        Lookup.Prep(assetCache);
+
+        ResponsesSpecific = LinkCache.Resolve<IDialogResponsesGetter>(FormKey.Factory("0D5533:Skyrim.esm"));
+        ResponsesGeneric = LinkCache.Resolve<IDialogResponsesGetter>(FormKey.Factory("0DBA2A:Skyrim.esm"));
```
Noggog, line 54: Just want to confirm these benchmark tests are just using skyrim base master content and don't rely on other mods you might have installed to measure properly

Mutagen.Bethesda.Skyrim/Records/Assets/VoiceType/VoiceTypeAssetLookup.cs
```diff
@@ -807,156 +815,123 @@ private IEnumerable<string> GetVoiceTypes(FormKey speaker)
...
-        if ((npc.Configuration.TemplateFlags & NpcConfiguration.TemplateFlag.Traits) == 0 && !npc.Race.IsNull)
-        {
-            return new HashSet<FormKey> { npc.Race.FormKey };
-        }
+        // TODO: Account for templates
+        if (!npc.Template.IsNull && npc.Configuration.TemplateFlags.HasFlag(NpcConfiguration.TemplateFlag.Traits))
+            return [];
 
-        return [];
+        return [npc.Race];
```
Noggog, line 934: Looked like the lhs side checked if the npc.Race was null first?  Is that removal intentional?  If so, maybe we update some of the unit tests to reflect

Might be good to brush up on those in general.  I remember at the time of making this (or maybe some similar asset logic) didn't have as many tests as desired.   Could be good to lock in the behavior with the tests first, then apply the optimization logic

Mutagen.Bethesda.Skyrim/Records/Assets/VoiceType/VoiceTypeAssetLookup.cs
```diff
@@ -807,156 +815,123 @@ private IEnumerable<string> GetVoiceTypes(FormKey speaker)
...
     {
-        if ((npc.Configuration.TemplateFlags & NpcConfiguration.TemplateFlag.Stats) == 0 && !npc.Class.IsNull)
-        {
-            return new HashSet<FormKey> { npc.Class.FormKey };
-        }
+        if (!npc.Template.IsNull && npc.Configuration.TemplateFlags.HasFlag(NpcConfiguration.TemplateFlag.Stats))
+            return GetClasses(npc.Template);
 
-        return npc.Template.IsNull ? new HashSet<FormKey>() : GetClasses(npc.Template);
+        return [npc.Class];
```
Noggog, line 895: As mentioned in the other comment, this seems to have logic differences.  

It feels like the old logic returned an empty set if the class and template was null.  Whereas this will return npc.Class even if it's null?

I wouldn't be too worried, except no unit tests needed to be changed, which suggests the behavior wasn't really flexed by them too well in the first place

What IS the intention?  Is it correct that we return empty if the class is null or not?   I'm not familiar enough with the goals to know the answer... but solidifying the choice in unit tests will at least solidify the intention one way or the other

Mutagen.Bethesda.Skyrim/Records/Assets/VoiceType/VoiceTypeAssetLookup.cs
```diff
@@ -807,156 +815,123 @@ private IEnumerable<string> GetVoiceTypes(FormKey speaker)
     }
 
     #region Voice Parser
-    private HashSet<string> GetVoiceTypes(INpcGetter npc)
+    private IEnumerable<string> GetVoiceTypes(INpcGetter npc)
```
Noggog, line 818: I dont mind the HashSet return value here.  It's private.  And we are returning hash sets. And they are more "useful" than straight IEnumerable.

I think dropping the type back to IEnumerable is only interesting in public interfaces where we might change our minds later on what the backing collection is.  N/A for private functions

Noggog (approved):
Thanks for the PR!   Great to see things getting optimized.

My main gripe is mainly a past sin where we didn't finish off writing unit tests to go with the logic.   This makes followups like this harder as we have to evaluate if the improvements changed any behaviors.    Are there edge case differences introduced in the change?   Is it faster just because it's not actually doing the same job?  Hard to tell/prove.

I do see -some- tests in the VoiceTypeAssetLookupTestSkyrim.cs file.  But I also noticed some logic changes that suggest the tests aren't quite complete.   Maybe/hopefully some of the differences i spotted are non-issues.  But I think it's worth considering upgrading the tests as well while we're here.   Even defining the tests help communicate intent very well.  Do we INTEND for X to be returned when Y is true, etc.

kieranknowles1:
> Thanks for the PR! Great to see things getting optimized.
> 
> My main gripe is mainly a past sin where we didn't finish off writing unit tests to go with the logic. This makes followups like this harder as we have to evaluate if the improvements changed any behaviors. Are there edge case differences introduced in the change? Is it faster just because it's not actually doing the same job? Hard to tell/prove.
> 
> I do see -some- tests in the VoiceTypeAssetLookupTestSkyrim.cs file. But I also noticed some logic changes that suggest the tests aren't quite complete. Maybe/hopefully some of the differences i spotted are non-issues. But I think it's worth considering upgrading the tests as well while we're here. Even defining the tests help communicate intent very well. Do we INTEND for X to be returned when Y is true, etc.

The main intentional change is to only consider winning overrides. I.e., if a plugin changes an NPCs voice, then only include the new voice.

You're right that unit tests should be extended, it's probably smart to do this in a separate branch based on the original implementation.

Mutagen.Bethesda.Skyrim/Records/Assets/VoiceType/VoiceTypeAssetLookup.cs
kieranknowles1: Most non-generic NPCs have exactly one voice type, so using a HashSet seems overkill.

With the addition of _voiceSpeakers for GetIsVoiceType conditions _speakerVoices may be redundant. I'll have to do some more testing to see if an enumerable is better and if removing _speakerVoices is worth the memory savings. The other _fieldNPCs lookups would probably benefit from being lazy, but less so from a usage cache since they can be inherited.

Mutagen.Bethesda.Tests/Benchmarks/VoiceTypeAssetLookupBenchmark.cs
kieranknowles1: I'm using mod organiser, so my testing environment consists only of the base masters + free CC.

Mutagen.Bethesda.Skyrim/Records/Assets/VoiceType/VoiceTypeAssetLookup.cs
Noggog: Ah gotcha.  Yeah, i see now where the logic can sometimes return not a hash set

kieranknowles1:
Closing this as #681 covers much of the same areas and may be impractical to merge.

PR #681, opened 2026-08-10
Extend Test Cases for VoiceTypeAssetLookup and fix differences from runtime behaviour

kieranknowles1:
Intentional changes to match runtime behavior:
- Remove handling of `AllowDefaultDialog` flag, this is used by the CK as described on the wiki (https://ck.uesp.net/w/index.php?title=VoiceType&oldid=17877), but has no runtime effect
- Build lookups based on winning overrides only
- Remove epsilon handling when inverting conditions. No such system is used at runtime
- Support filtering on `HasKeyword` conditions
- Group voice types based on FormKey rather than editor ID
- Expose `GetVoices`

Issues resolved:
- Inverting a `GetIsVoiceType` condition would exclude all non default dialog voices, even if they were explicitly included by another condition.
- Voice file paths for responses defined in a different plugin to the topic would use the topics mod key rather than that of the response.

Supersedes #679, but does not include its optimizations to ease review.

### DIFF
--- a/Mutagen.Bethesda.Skyrim/Records/Assets/VoiceType/VoiceContainer.cs
+++ b/Mutagen.Bethesda.Skyrim/Records/Assets/VoiceType/VoiceContainer.cs
@@ -1,4 +1,4 @@
-﻿using System.Text;
+using System.Text;
 using Mutagen.Bethesda.Plugins;
 using Noggog;
 namespace Mutagen.Bethesda.Skyrim.Records.Assets.VoiceType;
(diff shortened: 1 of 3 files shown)


Noggog:
@Elscrux you'll be more informed on the intention of the logic and if the tests are outlining them well

Noggog (approved):
My approval is pretty loose.  No general stylistic problems I can see.  Mostly just up to @Elscrux and his thoughts

Elscrux:
The logical changes as listed in the description and as I see them in the code all seem good, thanks @kieranknowles1 
Could you also update the CK Wiki page with your investigation results please?

kieranknowles1:
> The logical changes as listed in the description and as I see them in the code all seem good, thanks @kieranknowles1 Could you also update the CK Wiki page with your investigation results please?

There's a couple more places I've identified would benefit from test cases, I'll mark this PR as ready for review once those are in and update the wiki.

Noggog:
New code seems to have borked the tests

EDIT.  Oh, should read comments before i comment

Noggog (approved):

PR #689, opened 2026-08-27
Add value equality to GenderedItem<T>

### LINKED ISSUE #685: GenderedItem<T> has no Equals/GetHashCode → generated record Equals is false for identical records (record.Equals(record.DeepCopy()) == false)
## Summary

`GenderedItem<T>` does not override `Equals`/`GetHashCode`, so every generated record `Equals` that compares a gendered field (`ARMA.Priority`, `ARMA.WeightSliderEnabled`, `ARMO.WorldModel`, `RACE`, `NPC_` …) falls back to reference equality on it and returns `false` for two records with identical content — including a record and its own `DeepCopy()`.

## Repro (Fallout4, 0.53.1 and `dev` @ c4bdf427e)

```csharp
new GenderedItem<byte>(1, 1).Equals(new GenderedItem<byte>(1, 1));   // false

var fk = FormKey.Factory("000800:Test.esp");
var a = new ArmorAddon(fk, Fallout4Release.Fallout4) { Priority = new GenderedItem<byte>(1, 1) };
a.Equals(a.DeepCopy());                                               // false
a.GetEqualsMask(a.DeepCopy(), EqualsMaskHelper.Include.OnlyFailures).Priority; // (true, true) — mask says equal
```

## Where

- `Mutagen.Bethesda.Core/Plugins/Records/GenderedItem.cs` — `GenderedItem<T>` has `Male`, `Female`, indexer, enumerator, `Print`; no equality members.
- Generated `ArmorAddonCommon.Equals`: `if (!Equals(lhs.Priority, rhs.Priority)) return false;` → `object.Equals` → reference compare.
- The mask path (`GenderedItem.EqualityMaskHelper`) compares contents, which is why `GetEqualsMask` disagrees with `Equals`.

## Observed impact

Parsing the same plugin twice and comparing records with `Equals` reports 97 Armor + 64 ArmorAddon (plus RACE/NPC_) records unequal in a 684-plugin survey, all byte-identical.

## Suggested fix

Implement `Equals(object)`, `Equals(GenderedItem<T>)` and `GetHashCode` on `GenderedItem<T>` using `EqualityComparer<T>.Default` on `Male`/`Female` (matching what `EqualityMaskHelper` already does), plus a unit test in `Mutagen.Bethesda.Core.UnitTests`. Happy to open the PR.

WhiskyTangoFawks:
GenderedItem<T> previously relied on reference equality (inherited from object), so generated record Equals()/GetEqualsMask() comparisons that delegate to it (e.g. ArmorAddon.Priority) returned false/mismatched for two distinct instances holding identical Male/Female values — including armorAddon.Equals(armorAddon.DeepCopy()).

Implements IEquatable<GenderedItem<T>>: Equals(object), Equals(GenderedItem<T>), and GetHashCode(), comparing Male/Female via EqualityComparer<T>.Default so it works for both value types and loqui-object T.

Adds Core unit tests (GenderedItemTests) covering equal/unequal members, hash consistency, loqui-like T, null members, and Equals(object) null/wrong-type cases; adds a Fallout4 records-level test (ArmorAddonEqualityTests) verifying DeepCopy equality and GetEqualsMask agreement.

Fixes Mutagen-Modding/Mutagen#685

### DIFF
--- a/Mutagen.Bethesda.Core.UnitTests/Plugins/Records/GenderedItemTests.cs
+++ b/Mutagen.Bethesda.Core.UnitTests/Plugins/Records/GenderedItemTests.cs
@@ -1,0 +1,70 @@
+using Mutagen.Bethesda.Plugins.Records;
+using Shouldly;
+using Xunit;
+
+namespace Mutagen.Bethesda.Core.UnitTests.Plugins.Records;
+
+public class GenderedItemTests
+{
+    private sealed record LoquiLike(int Value);
+
+    [Fact]
+    public void Equals_SameMembers_ReturnsTrue()
+    {
+        var g1 = new GenderedItem<byte>(1, 1);
+        var g2 = new GenderedItem<byte>(1, 1);
+        g1.Equals(g2).ShouldBeTrue();
+        g1.Equals((object)g2).ShouldBeTrue();
+    }
+
+    [Fact]
+    public void Equals_DifferentMale_ReturnsFalse()
+    {
+        var g1 = new GenderedItem<byte>(1, 1);
+        var g2 = new GenderedItem<byte>(2, 1);
+        g1.Equals(g2).ShouldBeFalse();
+    }
+
+    [Fact]
+    public void Equals_DifferentFemale_ReturnsFalse()
+    {
+        var g1 = new GenderedItem<byte>(1, 1);
+        var g2 = new GenderedItem<byte>(1, 2);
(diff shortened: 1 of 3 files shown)


WhiskyTangoFawks:
Withdrawing this: on reflection the fix modifies generated/emitter output, and we've concluded it's better not to mess with the generated code from outside. We'll handle record equality on our own side instead. Leaving #685 closed alongside for the same reason — thanks, and sorry for the noise.

PR #690, opened 2026-08-27
Fix Dict-field Equals to compare by key, matching GetEqualsMask (#686)

### LINKED ISSUE #686: Fallout4 Package.Data: generated Equals compares the dictionary as an ordered sequence; GetEqualsMask disagrees
## Summary

`Package.Data` (Fallout4) is a `Dictionary<sbyte, APackageData>`, but the generated `PackageCommon.Equals` compares it with `SequenceEqualNullable` — i.e. as an **ordered** sequence of `KeyValuePair`s — so two packages with identical entries inserted in a different order are unequal. `GetEqualsMask` uses `EqualsMaskHelper.DictEqualsHelper` (keyed) and reports no failures, so the two APIs disagree.

## Repro (Fallout4, 0.53.1 and `dev` @ c4bdf427e)

```csharp
var fk = FormKey.Factory("000801:Test.esp");
var a = new Package(fk, Fallout4Release.Fallout4);
a.Data[5] = new PackageDataBool { Data = true };
a.Data[2] = new PackageDataBool { Data = false };
var b = new Package(fk, Fallout4Release.Fallout4);
b.Data[2] = new PackageDataBool { Data = false };
b.Data[5] = new PackageDataBool { Data = true };

a.Equals(b);                                                          // false
a.GetEqualsMask(b, EqualsMaskHelper.Include.OnlyFailures).Data;       // null — no failure
```

## Why it shows up in practice

`PackageBinaryWriteTranslation` writes `Data` in key order (`data.OrderBy(k => k.Key)`), while the parser inserts in on-disk `UNAM` order. Any plugin whose `UNAM`s aren't already sorted (CK output frequently isn't — 24 packages in one real plugin) round-trips to a `Package` that is content-identical but `Equals`-unequal.

## Where

- `Package_Generated.cs` (Equals): `if (!lhs.Data.SequenceEqualNullable(rhs.Data)) return false;`
- `Package_Generated.cs` (mask): `ret.Data = EqualsMaskHelper.DictEqualsHelper(...)`
- Generator: `Dict` fields in the Equals emitter.

## Suggested fix

Generator-level: emit a keyed comparison for `Dict` fields in `Equals` (count equal + every key present with equal value), matching the mask helper. Happy to open the PR if you can point me at the emitter.

WhiskyTangoFawks:
Loqui.Generation.DictType_Typical/DictType_KeyedValue emit Equals for Dict-typed fields via SequenceEqualNullable, an order-sensitive comparison. GetEqualsMask already compares the same fields by key via EqualsMaskHelper.DictEqualsHelper/CacheEqualsHelper, so two records with identical Dict content in a different insertion order were Equals-unequal despite GetEqualsMask reporting them equal - a real round-trip scenario (e.g. Fallout4 Package.Data: writer emits keys sorted, parser inserts in on-disk UNAM order).

Mutagen.Bethesda.Generation/Fields/DictType.cs now overrides GenerateForEquals/GenerateEqualsSnippet (both public, non-sealed on Loqui.Generation.DictType) to reuse the same DictEqualsHelper/ CacheEqualsHelper runtime calls GetEqualsMask already uses, branching on DictMode (KeyValue vs. Group.RecordCache's KeyedValue) and on whether the value type is a LoquiType (needs a maskGetter) or a primitive (EqualityComparer<T>.Default.Equals per key). Nullable Dict fields (e.g. Race.BipedObjects/BipedObjectNames) get an explicit null-guard preserving SequenceEqualNullable's null-tolerance, since the helpers themselves don't accept null dictionaries.

Regenerated via Generator.All across all five games; diff is scoped to the emitter plus the touched Dict-field Equals bodies (Package.Data, Group.RecordCache, Class.Attributes/SkillWeights/StatWeights, PlayerSkills.SkillValues/SkillOffsets, Race.BipedObjects/ BipedObjectNames/Starting/Regen). Excluded from this commit: unrelated generator drift (RecordTriggerSpecs ACHR/REFR ordering, MergedCellBlock/ MergedCellSubBlock dedup) present in Cell_Generated.cs/ Worldspace_Generated.cs/*MultiModOverlay_Generated.cs even on unmodified dev - confirmed via a control regen with DictType.cs reverted, out of scope for this ticket.

New tests: PackageDataEqualsTests (Package.Data, DictMode.KeyValue) and GroupRecordCacheEqualsTests (Group.RecordCache, DictMode.KeyedValue) in Mutagen.Bethesda.UnitTests, each covering reordered-same-content equality and differing-value-under-same-key inequality, both checked against Equals and GetEqualsMask(...).All(...).

Fixes Mutagen-Modding/Mutagen#686

### DIFF
--- a/Mutagen.Bethesda.Fallout3/Records/Fallout3Group_Generated.cs
+++ b/Mutagen.Bethesda.Fallout3/Records/Fallout3Group_Generated.cs
@@ -1059,7 +1059,7 @@ namespace Mutagen.Bethesda.Fallout3
             }
             if ((equalsMask?.GetShouldTranslate((int)Fallout3Group_FieldIndex.RecordCache) ?? true))
             {
-                if (!lhs.RecordCache.SequenceEqualNullable(rhs.RecordCache)) return false;
+                if (!(EqualsMaskHelper.CacheEqualsHelper(lhs: lhs.RecordCache, rhs: rhs.RecordCache, maskGetter: (k, l, r) => l.GetEqualsMask(r, EqualsMaskHelper.Include.All), include: EqualsMaskHelper.Include.All)?.Overall ?? true)) return false;
             }
             return true;
         }
--- a/Mutagen.Bethesda.Fallout3/Records/Major Records/Class_Generated.cs
+++ b/Mutagen.Bethesda.Fallout3/Records/Major Records/Class_Generated.cs
@@ -1633,7 +1633,7 @@ namespace Mutagen.Bethesda.Fallout3
             }
             if ((equalsMask?.GetShouldTranslate((int)Class_FieldIndex.Attributes) ?? true))
             {
-                if (!lhs.Attributes.SequenceEqualNullable(rhs.Attributes)) return false;
+                if (!(EqualsMaskHelper.DictEqualsHelper(lhs: lhs.Attributes, rhs: rhs.Attributes, include: EqualsMaskHelper.Include.All)?.Overall ?? true)) return false;
             }
             return true;
         }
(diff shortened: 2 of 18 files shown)


WhiskyTangoFawks:
Withdrawing this: same reasoning as #689 — the fix touches generated emitter code and we've decided not to mess with generated code from outside the project. We'll compare on our own side instead.

Issue #687, opened 2026-08-27
Fallout4 MaterialSwap: FNAM form-version gate reads MetaData.FormVersion before MajorRecordParse sets it, so valid v131 records throw "All FNAM strings should be the same"


WhiskyTangoFawks:
## Summary

`MaterialSwapBinaryCreateTranslation.FillBinaryFNAMParsingCustom` (Fallout4) gates its legacy "all FNAM strings must match" logic on `frame.MetaData.FormVersion >= 112`, but it runs as part of `fillStructs`, and `PluginUtilityTranslation.MajorRecordParse` assigns `frame.MetaData.FormVersion = record.FormVersion` **after** `fillStructs` returns. The gate therefore always sees `null`/a stale value, the legacy path always runs, and any `MSWP` record with two differing `FNAM`s throws `MalformedDataException: All FNAM strings should be the same` — even at form version 131, where the second `FNAM` is the substitution's obsolete tree-folder field and may legitimately differ.

## Repro

Real plugin: *Clipboards to the BOS.esp* (Nexus), `MSWP 0100002C`, form version 131, subrecords `FNAM "materials"` then `FNAM ""` inside a substitution. `Fallout4Mod.CreateFromBinary` throws; xEdit loads it. Stack:

```
MalformedDataException: All FNAM strings should be the same
  at MaterialSwapBinaryCreateTranslation.FillBinaryFNAMParsingCustom(MutagenFrame, IMaterialSwapInternal)
  at MaterialSwapBinaryCreateTranslation.FillBinaryStructs(IMaterialSwapInternal, MutagenFrame)
  at PluginUtilityTranslation.MajorRecordParse[M](...)
```

Vanilla never trips it only because no vanilla `MSWP` (3,937 checked across Fallout4.esm + DLC) has two distinct `FNAM` strings.

## Where

- `Mutagen.Bethesda.Core/Plugins/Binary/Translations/PluginUtilityTranslation.cs` `MajorRecordParse`: `fillStructs(record, frame)` precedes `frame.MetaData.FormVersion = record.FormVersion;`
- `Mutagen.Bethesda.Fallout4/Records/Major Records/MaterialSwap.cs` line ~28: `if (frame.MetaData.FormVersion >= NewFormVersion) return;`

## Suggested fix

[...]

PR #691, opened 2026-08-27
Set FormVersion before struct-fill so MaterialSwap's FNAM gate works …

WhiskyTangoFawks:
…(Mutagen-Modding/Mutagen#687)

PluginUtilityTranslation.MajorRecordParse set frame.MetaData.FormVersion from record.FormVersion only after fillStructs returned, but MaterialSwap's FillBinaryFNAMParsingCustom (Fallout4) reads frame.MetaData.FormVersion from inside fillStructs, so it always saw null/stale and took the legacy same-string-required path, throwing MalformedDataException on well-formed v112+ MSWP records whose second FNAM is a substitution's obsolete tree-folder field. Peek the FormVersion directly off the record's header bytes (frame.Reader.GetMajorRecordHeader().FormVersion) before fillStructs runs, since record.FormVersion itself isn't populated until fillStructs does so.

The overlay factory path had a separate, independent bug with the same symptom: MaterialSwapBinaryOverlay's generated FillRecordType switch guard used stream.MetaData.FormVersion, which is never populated anywhere in the overlay-mode read path (confirmed via full-repo grep of assignment sites) - so a v112+ record's top-level FNAM was unconditionally mis-routed into Substitutions-list parsing instead of being captured as TreeFolder, corrupting both fields for every overlay-mode read of a v112+ MaterialSwap, not just the malformed-duplicate-FNAM case. Root cause traced to the generator itself (PluginTranslationModule.cs:1754): the duplicate-trigger+versioning overlay guard emitted "stream.MetaData.FormVersion", while the sibling RecordTypeVersioning guard six lines above already used the correct, always-live "this._package.FormVersion!.FormVersion" pattern (matching BodyTemplate.cs's and ImageSpace.cs's own hand-written overlay idiom). Fixed the generator to use the same correct pattern and regenerated (Fallout4 single-game generator, then Generator.All - confirmed via diff that only MaterialSwap_Generated.cs changed; Generator.All's Cell/Worldspace trigger-order and MultiModOverlay cell-merging diffs across all five games are pre-existing drift between checked-in output and the generator, unrelated to this change, and were reverted). MaterialSwapBinaryOverlay's hand-written custom fill (MaterialSwap.cs) now reads this.FormVersion (the overlay's own live, on-demand header read) instead of the broken tracker.

Audit for other MetaData.FormVersion reads inside a *StructsCustom/ FillBinaryStructs path (the same deep-parse ordering bug would apply to any of them): none found. Every other FormVersion-gated field across Fallout3/Fallout4/Skyrim/Starfield (Race, Explosion, EffectShader, ImageSpace, CriticalData, Ammunition, Weather, LandscapeTexture, BodyTemplate, Leveled*, MagicEffectSound, WeaponDamageType, DamageTypeValue, ObjectProperty, BlockEditorMetaDataComponent) reads frame.MetaData.FormVersion only from within the per-subrecord dispatch loop (FillBinaryRecordTypes) or a write path, both of which already run after FormVersion is set. MaterialSwap.TreeFolder is the only field in the fork combining binary="Custom" with <Versioning> on a record type that collides with another field's trigger, so it was also the only one exercising the broken overlay-guard generator path.

Adds a fixture-free unit test (MaterialSwapFormVersionGateTests) that hand-builds minimal MSWP bytes in memory and covers both the deep parser and the overlay factory: a v131 record with two differing FNAMs now parses (TreeFolder from the top FNAM, substitution's own FNAM ignored), while a sub-112 record with differing FNAMs still throws MalformedDataException, preserving legacy behavior.

### DIFF
--- a/Mutagen.Bethesda.Core/Plugins/Binary/Translations/PluginUtilityTranslation.cs
+++ b/Mutagen.Bethesda.Core/Plugins/Binary/Translations/PluginUtilityTranslation.cs
@@ -88,6 +88,11 @@ internal static class PluginUtilityTranslation
         MajorRecordFill<M> fillTyped)
         where M : IMajorRecordGetter
     {
+        // Peek the record's FormVersion directly off its header bytes (without advancing
+        // the reader) so it is available to fillStructs' own custom struct-fill logic below.
+        // record.FormVersion isn't populated until fillStructs runs, so it can't be used yet;
+        // the header itself is already sitting right here to read from instead.
+        frame.MetaData.FormVersion = (ushort?)frame.Reader.GetMajorRecordHeader().FormVersion;
         frame = frame.SpawnWithFinalPosition(HeaderTranslation.ParseRecord(frame.Reader));
         fillStructs(
             record: record,
@@ -101,7 +106,6 @@ internal static class PluginUtilityTranslation
             }
 
             Dictionary<RecordType, int>? recordParseCount = null;
-            frame.MetaData.FormVersion = record.FormVersion;
             var lastParsed = new PreviousParse();
             while (!targetFrame.Complete)
             {
--- a/Mutagen.Bethesda.Core/Plugins/Binary/Translations/PluginUtilityTranslation.cs
+++ b/Mutagen.Bethesda.Core/Plugins/Binary/Translations/PluginUtilityTranslation.cs
@@ -88,10 +88,6 @@ internal static class PluginUtilityTranslation
         MajorRecordFill<M> fillTyped)
         where M : IMajorRecordGetter
     {
-        // Peek the record's FormVersion directly off its header bytes (without advancing
-        // the reader) so it is available to fillStructs' own custom struct-fill logic below.
-        // record.FormVersion isn't populated until fillStructs runs, so it can't be used yet;
-        // the header itself is already sitting right here to read from instead.
         frame.MetaData.FormVersion = (ushort?)frame.Reader.GetMajorRecordHeader().FormVersion;
         frame = frame.SpawnWithFinalPosition(HeaderTranslation.ParseRecord(frame.Reader));
         fillStructs(
--- a/Mutagen.Bethesda.Fallout4/Records/Major Records/MaterialSwap.cs
+++ b/Mutagen.Bethesda.Fallout4/Records/Major Records/MaterialSwap.cs
@@ -132,15 +132,12 @@ partial class MaterialSwapBinaryOverlay
 {
     private int? _fnamLoc;
     private int _offset;
-
+    
     public partial String? GetTreeFolderCustom()
     {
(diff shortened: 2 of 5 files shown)


Noggog:
Thanks for the PR and the deep dive into the bug!  Sorry for the late review.  I'm away from my large PC till after next week.  Will put this through the integration test ringer at that point, since it's touching pretty core mechanisms.  But looks like an improvement overall.  Will get back!

WhiskyTangoFawks:
No worries, if there are any changes needed let me know

Mutagen.Bethesda.Fallout4/Records/Major Records/MaterialSwap.cs
```diff
@@ -131,12 +131,14 @@ public static partial void WriteBinarySubstitutionsCustom(MutagenWriter writer,
...
+
     public partial String? GetTreeFolderCustom()
     {
-        if (_formVersion >= MaterialSwapBinaryCreateTranslation.NewFormVersion)
+        // this.FormVersion is read directly off this record's own header bytes, on demand -
+        // unlike _package.MetaData.FormVersion (a deep-parse-only tracker that is never
+        // populated for overlay reads), it is always correct here.
+        if (this.FormVersion >= MaterialSwapBinaryCreateTranslation.NewFormVersion)
         {
             return _fnamLoc.HasValue ? BinaryStringUtility.ProcessWholeToZString(HeaderTranslation.ExtractSubrecordMemory(_recordData, _fnamLoc.Value, _package.MetaData.Constants), encoding: _package.MetaData.Encodings.NonTranslated) : string.Empty;
```
Noggog: So i think there is a bug here too on this line, but it wasn't manifesting previously because of the bug you fixed.  But now that you have, it's expressing itself.   This should return `null` as the fallback.   Returning empty was writing FNAM stubs where there shouldnt have been.  

There are these [extra integration tests](https://mutagen-modding.github.io/Mutagen/Correctness/) that detected the issue.  They're a bit annoying to run since they need all the base esms present and take a long time. 

But anyway, can you change this to null before we merge in this PR?

Mutagen.Bethesda.Core/Plugins/Binary/Translations/PluginUtilityTranslation.cs
```diff
@@ -88,6 +88,11 @@ internal static M MajorRecordParse<M>(
         MajorRecordFill<M> fillTyped)
         where M : IMajorRecordGetter
     {
+        // Peek the record's FormVersion directly off its header bytes (without advancing
```
Noggog: Can we make the comments more minimal?  i typically like letting the code speak for itself when it can.  And/or letting the git history tell the story of reasons for changes to the file.

Noggog (changes requested):
Followup fix needed that the integration tests found

Noggog (approved):

PR #695, opened 2026-09-09
Derive DialogTopic Subtype and Category from SNAM

### LINKED ISSUE #590: DialogTopic implicitly connected fields SNAM, SubtypeID, Category
- [ ] In DialogTopic, the Subtype ID and Category fields should implicitly set by the SNAM field when reading a DialogTopic
- [ ] When editing a DialogTopic, all users should need to set is the Subtype ID which has an enum which is the main thing you'd want to interact with
- [ ] When writing a DialogTopic, the Subtype ID should be implicitly setting the Category and SNAM fields based on the Subtype ID

Avick3110:
Closes #590.

`DialogTopic.Subtype` currently mis-labels 1,474 of the 19,557 DIAL records the vanilla Skyrim
masters define (7.5%), because it reads `DATA\Subtype` and that field carries two incompatible
numberings.

### The renumbering

Bethesda inserted the six `FlyingMount*` subtypes at index 20 with the Dragonborn-era Creation Kit, so
every DIAL authored before that stores a `DATA\Subtype` six lower than `SubtypeEnum`'s current
numbering: `DIAL 0002707A:Skyrim.esm` has `DATA = 00 07 49 00` (73) beside `SNAM = HELO` (Hello = 79)
and reads back as `RechargeExit`, and `DIAL 000904AC` has 72 beside `GBYE` and reads back as
`Recharge` — `DATA` is a full four bytes on both, so this is not a truncated subrecord. Across the
15,037 DIAL records `Skyrim.esm` defines there are exactly two buckets and no third: delta 0 for every
record below index 20, and delta +6 for the 1,007 records at index 26 or above. Form version cannot
pick the table — `Dawnguard.esm`'s FormVersion-43 topics use the *old* numbering while
`Dragonborn.esm` mixes both inside one plugin at FormVersion 43, plus 38 records at delta +1 or −5
that Bethesda simply never maintained.

Scan over all 19,557 DIAL records the five vanilla masters define:

| Plugin | DIAL defined | at delta 0 | at delta +6 | at other nonzero delta | category byte alone stale | DATA bytes rewritten by this change |
|---|---|---|---|---|---|---|
| Skyrim.esm | 15,037 | 14,030 | 1,007 | 0 | 0 | 1,007 |
| Update.esm | 31 | 31 | 0 | 0 | 0 | 0 |
| Dawnguard.esm | 1,936 | 1,649 | 287 | 0 | 2 | 289 |
| HearthFires.esm | 395 | 395 | 0 | 0 | 0 | 0 |
| Dragonborn.esm | 2,158 | 1,940 | 180 | 38 | 19 | 237 |
| **Total** | **19,557** | **18,045** | **1,474** | **38** | **21** | **1,533** |

The last column is the sum of the three before it: every record at a nonzero delta, plus the 21 `SCEN`
topics whose subtype already agrees with SNAM but whose category byte reads `Topic` instead of `Scene`.
Also from the same scan: not one vanilla DIAL has a blank or zero SNAM, and not one carries a marker
outside the marker table in `DialogTopic.cs`, so the default branch never fires on vanilla data.

Nothing on the record distinguishes the two numberings, so SNAM is the only reliable statement of a
topic's subtype. That is xEdit's position too: it marks `DATA\Subtype` `cpIgnore` and drives it from
SNAM (`wbDefinitionsTES5.pas`, `wbRecord(DIAL, …)`).

### What changed

SNAM drives a topic's subtype, and DATA's category and subtype bytes only restate it, so a record
whose halves disagree is corrupt rather than meaningful. There is therefore one exposed field, and
the rest is read and written under the hood.

`DialogTopic.xml` — `Category` and `SubtypeName` are removed as fields. `TopicFlags` stays, as the
only real data DATA carries, and `Subtype` is the single exposed enum. DATA and SNAM are handled by
custom logic with no fields behind them. `DialogTopic_Generated.cs` regenerated; the generator run
touched no other record.

`DialogTopic.cs` — adds the `SubtypeEnum` ↔ SNAM marker ↔ `CategoryEnum` table, the lookups over it,
and hand-written parse, write and overlay logic. The markers come from xEdit's `wbSubtypeNamesEnum`.

- **Read.** `Subtype` comes from the SNAM marker. DATA's category and subtype bytes are not read at
  all: nothing compares them against SNAM and nothing throws on a disagreement, since reporting
  corrupt content belongs in a tool rather than in the library. A SNAM that is absent, shorter than
  four bytes, or names a marker the enum does not model leaves `Subtype` at its default, `Custom`.
  `TopicFlags` is still read from DATA, length-checked.
- **Write.** `Subtype` decides DATA's category and subtype bytes and the SNAM marker, so a caller only
  ever sets one thing. A `Subtype` cast from an int that the enum does not model has no marker and no
  category, so it writes a zeroed SNAM rather than throwing.
- **Overlay.** The generated accessors indexed from the DATA range's start without checking the
  subrecord's length, so a short DATA read past it; the hand-written accessor is bounded to the
  subrecord, and a SNAM shorter than four bytes is treated as no marker.

`SkyrimProcessor.ProcessDialogs` derives the expected passthrough bytes the way the writer derives
them, for every shape it can substitute into: a DATA of four bytes or more beside a SNAM of four bytes
or more takes the subtype the marker names, and anything else expects the default subtype and its
category. Shapes that would change a subrecord's length are not substituted; no vanilla DIAL has one.

Skyrim only, for now. Fallout 4 and Starfield `DialogTopic` have the same DATA/SNAM shape in Mutagen —
Starfield already models SNAM as `SubtypeNameEnum` — and I started with Skyrim because that is where
the renumbering evidence is. Happy to follow with the other two.

### Behaviour change for existing consumers

- `DialogTopic.Category` and `DialogTopic.SubtypeName` no longer exist. Code that read `Category`
  should read `Subtype`: each category is a contiguous block of subtypes. `CategoryFromSubtype` is
  public if the byte itself is wanted — the passthrough processor uses it — and `SubtypeFromMarker`
  is public too.
- `Subtype` no longer round-trips the raw DATA bytes. On a pre-Dragonborn topic it now reports what
  SNAM says (`Hello`, not `RechargeExit`). Code that read `Subtype` for its numeric value gets a
  different — correct — answer. A topic with no usable SNAM reads as `Custom` rather than as DATA's
  number.
- Writing an unmodified vanilla topic back out is no longer byte-identical: DATA's category and
  subtype are renumbered to the modern table to match SNAM. That is 1,533 records across the vanilla
  masters (Skyrim.esm 1,007, Dawnguard.esm 289, Dragonborn.esm 237).

  I am not claiming the tools do this. xEdit only rewrites `DATA\Subtype` when SNAM is edited (its
  `SetAfterSet`); a copy-as-override is byte-verbatim. And `Dragonborn.esm` carries 180
  legacy-numbered records inside a plugin the Dragonborn-era Creation Kit saved, so the CK does not
  renumber on save either. The case for normalising here is narrower. One, the processor already does
  exactly this kind of fix: the "Reset misnumbered counter" block sits directly below the new code and
  rewrites Bethesda's stale TIFC counts on passthrough. Two, the engine has to be reading SNAM, since
  one plugin carries both numberings and both play in game — so a record whose DATA disagrees with its
  SNAM is stale data, not a second meaning.
- `SubtypeEnum` gains `CustomFVDL = 103`. `FVDL` is the one marker the enum did not name — xEdit
  labels it "Custom?" and it appears in no vanilla master. With `SubtypeName` gone there is no field
  left to carry it verbatim, so it gets a member, named as Starfield's `SubtypeNameEnum` already names
  that marker. Its `Misc` category comes from where 103 sits in the enum's category blocks rather than
  from any vanilla record, and the writer and the processor read it from the same table. A topic using
  `FVDL` keeps its marker on write and has DATA's subtype bytes rewritten to 103.
- A default-constructed topic writes `SNAM = CUST` with `DATA = 00 00 00 00`, where it previously
  wrote four zero bytes and an empty SNAM.
- `VoiceTypeAssetLookup` (`Records/Assets/VoiceType/VoiceTypeAssetLookup.cs`) asked for the `Scene`
  category and now asks for the `Scene` subtype, the only member of that category. On a
  legacy-numbered topic its `SharedInfo` comparison previously missed and now matches — a favourable
  change in shipped voice-asset lookup, but a change.

### Two things worth a maintainer's call

1. **The Subtype → Category map has no upstream source, and it is the weakest-evidence part of this
   change.** Mutagen has no such table and xEdit does not publish one. I derived it from the vanilla
   masters (Skyrim, Update, Dawnguard, HearthFires, Dragonborn): for each SNAM marker, the
   `DATA\Category` byte those records carry. Of the markers that appear there, all but four carry one
   category throughout. The four have strays: `DETH` 1 of 84, `HELO` 1 of 406, `HIT_` 1 of 90, `SCEN`
   21 of 8,784 — every stray reads `Topic`, and the majority value is used.

   Nineteen markers appear in no vanilla master: `ASKF`, `ASKG`, `FAVO`, `FLAT`, `FOLL`, `FRJT`,
   `GIFF`, `JUMP`, `PCPS`, `PCSS`, `RCEX`, `RECH`, `REEX`, `REPA`, `SERU`, `SHRE`, `TRAI`, `TRAV`,
   `TREX`, and `FVDL` makes twenty. Each takes its enum block's category. That is not twenty separate
   guesses: `SubtypeEnum` is laid out in contiguous category blocks — Topic 0–2, Favor 4–13, Scene 14,
   Favors 15–25, Combat 26–54, Detection 55–65, Service 66–74, Misc 75–103 — and every marker vanilla
   *does* use falls in its block, with no crossings. So it is one structural inference, checked
   against the whole corpus.

   Still, this is the part I would most expect you to want a second look at.
2. **Passthrough was not run here.** The `SkyrimProcessor` change is reasoned, not verified against
   real game data — the piece most likely to need a second pass on your side.

### How tested

Added `DialogTopicSubtypeTests` — 29 cases; every read case exercises both the direct parser and the
overlay:

- the legacy reproducers above, on raw record bytes
- a modern-numbered record that already agrees
- SNAM before DATA, and SNAM parted from DATA by another subrecord, both also checking `TopicFlags`
  still comes from DATA
- the absent, short and unmodeled-marker SNAM cases, which leave `Subtype` at `Custom`
- a two-byte DATA that must not over-read; no DATA at all; a zero-length DATA
- a write that sets only `Subtype` and produces `DATA = 00 07 4F 00` with `SNAM = HELO`
- a default-constructed topic, and a `Subtype` cast from an int the enum does not model
- a legacy record round-tripped back out renumbered, and an `FVDL` record round-tripped through
  `CustomFVDL`, both checking the overlay writes the same bytes as the direct parser
- `CopyInFromBinary` of a `GBYE` record over a topic that already held `Hello`
- a table round-trip over every `SubtypeEnum` value

`dotnet test Mutagen.Records.sln`: 4,164 passed, 0 failed, 4 skipped
(`Mutagen.Bethesda.UnitTests` 3,373, `Mutagen.Bethesda.Core.UnitTests` 789,
`Mutagen.Bethesda.WPF.UnitTests` 2), the same on both `net9.0` and `net10.0`.
`Mutagen.Bethesda.Tests` builds clean on both. Passthrough tests against real game data were not run.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


Noggog (changes requested):
Heya!  Thanks so much for the hard work on the PR!

I think we should adjust a bit further.  SNAM seems to "drive" the category/Subtype fields' content, or something similar?   They should never deviate from each other, as that would be corrupt content?

If so, we should just have a single exposed field, full stop.  If the other content is derivative/implicit, it shouldn't even be exposed as fields to tweak.   This logic sets those fields properly.. but then leaves them open for corrupting tinkers after.

I think we just expose the SubType enum as a field, and then read/write the other data under the hood without defining fields for them.   

The read side can honestly just probably skip the bytes since we can reconstitute them when writing implicitly?   No error checking on read.   That's more the privvy of https://github.com/Mutagen-Modding/Antigen to detect corruption and complain, rather than code here in mutagen fundamentally throwing on disagreements which makes it harder to use.

Avick3110:
That makes sense, and you are right that the two can only ever disagree by being corrupt.
Reworked: `Category` and `SubtypeName` are gone as fields, `Subtype` is the only thing
exposed, and DATA's category and subtype bytes are no longer read at all — the subtype comes
from SNAM, and both bytes are written back from it. A SNAM that is absent, too short, or
carries a marker we do not model just leaves `Subtype` at its default; nothing checks the two
against each other and nothing throws.

One knock-on: with `SubtypeName` gone there was nothing left to carry `FVDL`, so `SubtypeEnum`
has a `CustomFVDL` member now, named the way Starfield's `SubtypeNameEnum` already names that
marker, and no vanilla master uses it. The subtype to category lookup stays public because the
passthrough processor needs it to work out the bytes a record should write. Full suite passes
on net9.0 and net10.0, but I can't run passthrough here.

(Hope you don't mind my agent handling my comments and replies, just letting you know there is a real human here!)

Mutagen.Bethesda.Tests/Processing/SkyrimProcessor.cs
```diff
@@ -235,6 +235,29 @@ private void ProcessDialogs(
             formKey,
             fileOffset);
 
+        // Derive DATA's category and subtype from SNAM
+        if (majorFrame.TryFindSubrecord(RecordTypes.DATA, out var dialDataRec)
```
Noggog: Hmm.  I don't think we should need to mutate the existing skyrim data in order to make the passthrough tests complete?  

Part of the test of correctness ON our modified setup would be that the bytes matched.   By mutating them here, we're losing that confirmation.

